PR #80, opened 2015-05-07
WIP: Open-source microsoft.fx.portability.offline

mjrousos:
This isn't ready to merge yet, but I'm sending it out to start the conversation about what it looks like to open-source the following:
* Microsoft.FX.Portability.Offline
* The HTML and Json report writers
* The reviewed breaking change markdown files

I believe that these are all more-or-less ready to open-source. The only outstanding issue is reviewing catalog.bin and getting it into a proper home (notice the "TBD" in download.ps1 for this).

I think it will be useful for our customers to be able to get all of the ApiPort components they need to run locally from our repo, rather than needing us to fill in the gaps. This set of changes (and open-sourcing catalog.bin), should get us to that state.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ rcf
 *.swp
 .vs
 project.lock.json
+src/Microsoft.Fx.Portability.Offline/data/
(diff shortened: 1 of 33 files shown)


mjrousos:
Closed because the head branch was disconnected from this branch as part of making the repo public. Will create a new PR shortly.

PR #85, opened 2015-07-08
Add offline execution via Microsoft.Fx.Portability.Offline

mjrousos:
These changes add the following (optional) components to ApiPort -

* Microsoft.FX.Portability.Offline (for offline analysis)
* A few seed breaking change files to allow for offline breaking change analysis (the expectation is that more breaking change files will be added as UE and subject matter experts review them)
* The HTML and Json report writers for offline use (including a couple minor bug fixes).

With these changes, users can inject offline components (via Unity.config) and run ApiPort entirely offline.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ rcf
 *.swp
 .vs
 project.lock.json
+src/Microsoft.Fx.Portability.Offline/data/
--- a/Documentation/BreakingChanges/! Template.md
+++ b/Documentation/BreakingChanges/! Template.md
@@ -1,0 +1,34 @@
+## ID: Breaking Change Title
+
+### Scope
+Major
+
+### Version Introduced
+4.5
+
+### Version Reverted
+4.6
+
+### Change Description
+Description goes here.
+
+- [ ] Quirked
+- [ ] Build-time break
+- [ ] Source analyzer available
+
+### Recommended Action
+Suggested steps if user is affected (such as work arounds or code fixes) go here.
+
+### Affected APIs
+* Not detectable via API analysis
+
+[More information](LinkForMoreInformation)
+
+<!--
+    ### Original Bug
+    Bug link goes here
+    ### Notes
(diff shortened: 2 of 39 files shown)


twsouthwick:
It's a pretty big review, so if there's anything you want a careful review of, let me know.  I looked through and didn't see anything that stood out to me, so looks good.

mjrousos:
Yea, the review is big since I'm porting some components that were previously built internally. Probably the best way to look at it is to look at individual commits and skip the first and third ones (since those are just copying already-existent assets wholesale).

twsouthwick:
That helps a lot!  Why are we downloading the breaking changes since they are in the same repo?  Can't we just use the ones already in the repo?

mjrousos:
Good point :) There was a time when they weren't in this repo, but I decided to just move them. I'll update things to use those instead of re-downloading from Github.

twsouthwick:
I would prefer embedding the md files instead of having them next to it so that the offline library can be self contained

mjrousos:
Yea, I can see the value of the library being self-contained. On the other hand, though, I think it's important the users be able to override which breaking changes they're looking for (either to remove some they don't care about or add others).

Perhaps I can embed them *and* drop them next to the library (so that it's obvious that's an option) and update the library to fall back to embedded ones if none are found externally?

twsouthwick:
That's what I've done in the past - that way the library itself is functional, but can be augmented with new info easily.  If two with the same ID are available, we should prefer the one in the directory (ie it supersedes the embedded version)

mjrousos:
Ok, I've updated the PR to work that way.

twsouthwick:
Looks like it will only take one or the other.  That's fine, but we may want to look at being selective (like I commented above) at some point.  Looks good to me!

mjrousos:
I made it just use one or the other intentionally, that way users can easily remove breaking changes from the set of what's scanned for as well as add. As you say, we can look at adding selectively if that becomes an important scenario (wouldn't be too big of a change).

PR #87, opened 2015-07-13
Fix resource for targetMap.xml

twsouthwick:
There were some issues with resources when a target map is available.  This fixes that


conniey:
LGTM

PR #88, opened 2015-07-13
Update tests to use new roslyn and metadata reader

twsouthwick:
This also include tests that were skipped before due to needing a newer System.Reflection.Metadata.  Since this is available now, the skip attribute is removed.


tests/Microsoft.Fx.Portability.MetadataReader.Tests/ManagedMetadataReaderTests.cs
```diff
@@ -113,10 +115,15 @@ private void CompareSpecificDependency(string path, string v)
...
 
-            var found = dependencies.Dependencies
-                .Any(d => string.Equals(d.Key.MemberDocId, v, StringComparison.Ordinal));
+            foreach (var dependency in dependencies.Dependencies)
+            {
+                if (string.Equals(dependency.Key.MemberDocId, v, StringComparison.Ordinal))
+                {
+                    return;
+                }
+            }
```
mjrousos: Why the change here? Just a perf optimization?

tests/Microsoft.Fx.Portability.MetadataReader.Tests/ManagedMetadataReaderTests.cs
```diff
@@ -137,7 +144,16 @@ private void CompareDependencies(string path, IEnumerable<Tuple<string, int>> ex
...
+            Assert.Equal(expectedOrdered.Count, foundDocIds.Count);
+
+            foreach (var combined in expectedOrdered.Zip(foundDocIds, Tuple.Create))
+            {
+                var expectedItem = combined.Item1;
+                var actualItem = combined.Item2;
+
+                Assert.Equal(expectedItem.Item1, actualItem.Item1);
+                Assert.Equal(expectedItem.Item2, actualItem.Item2);
+            }
```
mjrousos: I'd have hoped that Assert.Equal would have done the same thing as this code already. I take it that isn't the case.

conniey:
LGTM

mjrousos:
A couple comments regarding things I'm curious about, but LGTM.

tests/Microsoft.Fx.Portability.MetadataReader.Tests/ManagedMetadataReaderTests.cs
twsouthwick: I think I did it for debugging purposes - I kept changing it to this to identify what the dependencies are.  FIgured it would be better to keep it.

tests/Microsoft.Fx.Portability.MetadataReader.Tests/ManagedMetadataReaderTests.cs
twsouthwick: Again, this is to help with debugging when something fails - I found that the Assert.Equal made it difficult to test to find what the failing issue was.

mjrousos: Ah, I see. Makes sense.

tests/Microsoft.Fx.Portability.MetadataReader.Tests/ManagedMetadataReaderTests.cs
mjrousos: I see.

mjrousos:
LGTM

PR #89, opened 2015-07-13
Update build scripts

conniey:
Fixing the build scripts to install the latest dnx version when one does not exist.


mjrousos:
LGTM

PR #92, opened 2015-07-15
Make sure that the member name contains only safe characters.

### LINKED ISSUE #43: Iterator block for IEnumerable<T1<T2,T3>> generates incorrect docid
Use the following as the input assembly:

	using System;
	using System.Collections.Generic;

	namespace Microsoft.Fx.Portability.MetadataReader.Tests
	{
		class CallOtherClass
		{
			static void Main(string[] args)
			{
				OtherClass.GetValues<int>();
			}
		}

		public class OtherClass
		{
			internal static IEnumerable<Tuple<T, int>> GetValues<T>()
			{
				yield break;
			}
		}
	}

The docid generated for GetValues<T> is incorrect - there is a `@` where there should be a `#`.
	
**Expected:** ``M:Microsoft.Fx.Portability.MetadataReader.Tests.OtherClass.<GetValues>d__0`1.System#Collections#Generic#IEnumerable{System#Tuple{T@System#Int32}}#GetEnumerator``
**Actual:** ``M:Microsoft.Fx.Portability.MetadataReader.Tests.OtherClass.<GetValues>d__0`1.System#Collections#Generic#IEnumerable{System#Tuple{T,System#Int32}}#GetEnumerator``

The generics may be a red herring, but that's the only way I can get it to compile in a way that it is detected by the dependency finder.  There are probably compiler optimizations in other cases.

AlexGhiondea:
When we are producing the DocId for a member we need to make sure that name only contains safe characters.
This change introduces a method that does the following replacements:
'<' -> '{'
'>' -> '}'
'.' -> '#'
',' -> '@'

This method will be used for all the members on types.
Fixes #43.

### DIFF
--- a/src/Microsoft.Fx.Portability.MetadataReader/MemberMetadataInfo.cs
+++ b/src/Microsoft.Fx.Portability.MetadataReader/MemberMetadataInfo.cs
@@ -221,6 +221,26 @@ namespace Microsoft.Fx.Portability.Analyzer
             }
         }
 
+        /// <summary>
+        /// Replace the non-safe characters (<>.,) in a method name with safe ones ({}#@)
+        /// </summary>
+        /// <param name="name"></param>
+        /// <returns></returns>
+        private string GetDocIdSafeMemberName(string name)
+        {
+            char[] newName = new char[name.Length];
+            for (int i = 0; i < name.Length; i++)
+            {
+                if (name[i] == '<') newName[i] = '{';
+                else if (name[i] == '>') newName[i] = '}';
+                else if (name[i] == '.') newName[i] = '#';
+                else if (name[i] == ',') newName[i] = '@';
+                else newName[i] = name[i];
+
+            }
+            return new string(newName);
+        }
+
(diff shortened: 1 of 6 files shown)


twsouthwick:
Thanks! Since this fixes issue #43, can you add a test case for that?

src/Microsoft.Fx.Portability.MetadataReader/MemberMetadataInfo.cs
```diff
@@ -221,6 +221,26 @@ private IEnumerable<string> GetGenericDisplayNames(IList<string> displayNames)
             }
         }
 
+        /// <summary>
+        /// Replace the non-safe characters in a method name with safe ones.
```
twsouthwick: Can you add a description of what "safe" characters are?

AlexGhiondea:
Done.

conniey:
LGTM

twsouthwick:
LGTM.  Thanks!

PR #93, opened 2015-07-15
Using official releases on nuget.org rather than the nightly builds

conniey:
This is so we consistently create a NuGet package with the same dependencies instead of them changing with the ASP.NET nightly builds.


twsouthwick:
LGTM

PR #97, opened 2015-07-17
Add tests based on docid spec

### LINKED ISSUE #91: Additional test coverage for docid
Looks like there is a spec for docid generation with some good examples: https://msdn.microsoft.com/en-us/library/fsbx0t7x.aspx.  We should use those to add some more test cases to ensure we are generating things correctly.

twsouthwick:
@AlexGhiondea found a spec on (msdn)[https://msdn.microsoft.com/en-us/library/fsbx0t7x.aspx] that has examples of docid.  This adds tests for those.  In the process of adding them, I also found that using xUnit theory based tests will make it much easier to add future tests and cleaner to understand what is happening.

I removed the BaselineTests as they are flaky (depend on specific mscorlib version) and we have a bunch of regression tests now.

Fixes #91


tests/Microsoft.Fx.Portability.MetadataReader.Tests/ReflectionMetadataInfoTests.cs
```diff
@@ -44,17 +44,16 @@ public void UnresolvedAssemblyTest()
 
         private IEnumerable<string> _expectedResult = new[]
         {
-            "xunit.assert, Version=2.1.0.2945, Culture=neutral, PublicKeyToken=8d05b1bb7a6fdb6c",
-            "NSubstitute, Version=1.8.1.0, Culture=neutral, PublicKeyToken=92dd2e9066daa5ca",
+            "Microsoft.CodeAnalysis, Version=1.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35",
```
AlexGhiondea: Does this mean that you'll have to update this every time we pick up a newer drop?

twsouthwick: Unfortunately, yes.  We could come up with a test that creates an assembly and doesn't use itself, but this was a simple route to take.

AlexGhiondea:
LGTM.

PR #98, opened 2015-07-17
Remove silverlight if/defs

twsouthwick:
These if/defs made it compile but it was never tested and we don't need this support.


mjrousos:
LGTM

PR #116, opened 2015-08-20
Update ApiPort to v1.1.0

twsouthwick:
We haven't updated the version of ApiPort even though now it supports breaking change information as well as portability analysis.  This bumps the version to 1.1.0, according to [SemVer 2.0](http://semver.org/).


mjrousos:
LGTM

jasonwilliams200OK:
:+1: 
Aside, according to semver, unless you reach 1.0 i.e. (0.x.x), you can make as many breaking changes as you want. Once you reach 1.0, then all semver rules are applied. So starting off with v0.1.0 is a wise idea. To wit, node.js latest version is v0.12.x.

twsouthwick:
Good to know - although we started off with 1.0 a year ago :( next project we'll have to keep that in mind!

PR #120, opened 2015-08-24
Update markdown parser to parse escaped code and rename 'Documentation' to 'docs'

twsouthwick:


src/Microsoft.Fx.Portability/BreakingChangeParser.cs
```diff
@@ -198,8 +198,8 @@ private static void ParseNonStateChange(BreakingChange currentBreak, ParseState
                     }
                     break;
                 case ParseState.AffectedAPIs:
-                    // Trim md list markers, as well as comment tags (in case the affected APIs section is followed by a comment)
-                    string api = currentLine.Trim().TrimStart('*', '-', ' ', '\t', '<', '!', '-');
+                    // Trim md list and code markers, as well as comment tags (in case the affected APIs section is followed by a comment)
+                    string api = currentLine.Trim().TrimStart('*', '-', '`', ' ', '\t', '<', '!', '-').TrimEnd('`');
```
mjrousos: This will specifically trim the code markers from items in the affected APIs list. Is that an expected scenario? 

I think this change seems fine, but I'm not sure why we need it.

PR #128, opened 2015-08-28
Implementation of file parsing as byte[] instead of using file path.

mayaganesan:


twsouthwick:
Thanks @mayaganesan! I like the idea this is going for. However, I don't want to add overloads for different ways we can pass file types in. The reason I chose a filestream was because it contained Path and Stream. I'd prefer some abstraction like this:

```csharp
public interface IDependencyFile
{
    public string Path {get;}
    public Task<Stream> GetContentsAsync();
}
```

That way, both FileInfo or byte[] objects could be passed in, and have one API for it. What do you think?

mayaganesan:
Sure, a Stream should work fine. I chose to go with a byte array because there seems to be a read/write timeout in some Streams and I didn't want to run into any issues there. But all the places that require file paths should be modified in this pull request, so you can go ahead and make the change if you think that would be better.

conniey:
Just saw this old PR and was wondering what the solution should be? Is this what you had in mind?
```csharp
IDependencyInfo FindDependencies(IEnumerable<Stream> inputStreams, IProgressReporter progressReport);
```
or
```csharp
IDependencyInfo FindDependencies(IEnumerable<byte[]> input, IProgressReporter progressReport);
```
The latter would remove the dependency on us to use a File.Open in MetadataDependencyFinder.

twsouthwick:
The problem with passing a Stream is that it requires it to be open. It would be better to have some type (an IFileInput) that has a method like (Stream OpenRead()) that will open a file to be read (maybe an async version?). Then a byte[], fileinfo, or other type could be implemented.

twsouthwick:
I'm going to close this as I've done some work with this in another branch and will open a new PR. Please see #201 for that work.

PR #131, opened 2015-09-01
Add breaking changes 60-69

twsouthwick:


docs/BreakingChanges/065 Corner case WPF bug in which databound text box fails is fixed.md
```diff
@@ -0,0 +1,22 @@
+## 65: Corner case WPF bug in which databound text box fails to update is fixed
+
+### Scope
+Transparent
+
+### Version Introduced
+4.5
+
+### Change Description
+TODO
```
mjrousos: This breaking change has some 'TODO's for description and recommendation which ought to be filled in.

docs/BreakingChanges/065 Corner case WPF bug in which databound text box fails is fixed.md
```diff
@@ -0,0 +1,22 @@
...
+- [ ] Build-time break
+- [ ] Source analyzer planned
+
+### Recommended Action
+TODO
+
+### Affected APIs
+* Not detectable via API analysis
+
+[More information](https://social.msdn.microsoft.com/Forums/vstudio/en-US/c05a8c02-de67-47a9-b4ed-fd8b622a7e4a/if-i-target-net-40-but-run-on-a-machine-that-has-net-45-will-net-40-wpf-bugs-still-be-there?forum=wpf)
```
mjrousos: This link references the break, but also is a forum post with tons of unrelated 4.5 compat complaining. We should see if we can find a more targeted/useful link.

docs/BreakingChanges/068 DbParameter.Precision and DbParameter.Scale are now public virtual members.md
```diff
@@ -0,0 +1,32 @@
...
+* `M:System.Data.Common.DbParameter.set_Precision(System.Byte)`
+* `P:System.Data.Common.DbParameter.Scale`
+* `M:System.Data.Common.DbParameter.get_Scale`
+* `M:System.Data.Common.DbParameter.set_Scale(System.Byte)`
+
+[More information](https://msdn.microsoft.com/en-us/library/dn458356(v=vs.110).aspx#ADO)
+
+<!--
+    ### Notes
+    We can probably find this. But do we want to bother?
```
mjrousos: Since we're planning to do the analyzer at this point, we can probably lose the out-dated note here.

PR #138, opened 2015-09-02
Add breaking change documents 120-129

twsouthwick:


docs/BreakingChanges/126 GZipStream does not flush until it is disposed.md
```diff
@@ -0,0 +1,30 @@
...
+- [ ] Quirked
+- [ ] Build-time break
+
+### Recommended Action
+When using a GZipStream to write to a MemoryStream or other stream that will be read from within the same process, make sure that the GZipStream has been disposed before attempting to read from the output stream. Note that calling `Flush` is insufficient. Because the GZipStream is being disposed, it may be necessary (depending on specific code patterns) to use the GZipStream constructor that takes a boolean specifying to leave its underlying stream open when the GZipStream object is disposed.
+
+### Affected APIs
+* `M:System.IO.Compression.GZipStream.Flush`
+
+[More information](http://www.mistergoodcat.com/post/Working-around-a-bug-in-MicrosoftBclCompression)
```
mjrousos: This is an interesting break because it's the only one (I think) that has a more info link pointing somewhere other than MSDN, Connect, or SO. I think that's ok - the blog we link here does describe the issue and workaround pretty well - but there's a danger the site could vanish in the future.

Probably, we should have some automation to make sure that More Info links are valid and make a work item to create an MS-owned blog post about this issue (so that we can stop pointing at MisterGoodCat.com)

mjrousos: I chatted with Shawn. He recommended getting Jamshed's input on this.

twsouthwick: I've remove this one for now, then. We can add it after we sort that out.

PR #145, opened 2015-09-16
Show better help message when tool is run

twsouthwick:
The current help message is not very informative and seems to cause confusion. This updates the help messages to show more detailed info:

```
Microsoft (R) API Portability Analyzer v1.1.0
Copyright (C) Microsoft Corporation. All rights reserved.

This tool analyzes .NET assemblies to determine possible problems moving between .NET platforms (such as Windows Store, desktop, Mono, .NET Core, etc) as well as between .NET Framework versions (ie 4.x->4.y).

To learn more about how this tool works, including the data we are collecting, go here - http://go.microsoft.com/fwlink/?LinkId=397652

====================================================================================================
ApiPort.exe analyze [options]

Analyzes specified files/directories for IL and determines the APIs that are used. This information is analyzed for portability concerns or known breaking changes (depending on input parameters) and returns a report that gives a summary of issues and possible steps forward.

  -e, --endpoint=VALUE       Service endpoint
  -f, --file=VALUE           [Required] Path to assembly file or directory of
                               assemblies.
  -o, --out=VALUE            Output file name
  -d, --description=VALUE    Description of the submission
  -t, --target=VALUE         The target you want to check against.
  -r, --resultFormat=VALUE   The report output format
  -p, --showNonPortableApis  Calculate non-portable APIs
  -b, --showBreakingChanges  Calculate breaking changes on full .NET Framework
      --noDefaultIgnoreFile  Do not use the standard assembly ignore list
                               when analyzing breaking changes. The default
                               ignore list can be found at KnownSafeBreaks.json
  -i, --ignoreAssemblyFile=VALUE
                             Specifies a json file defining assemblies that
                               should not be analyzed for specific targets
                               while analyzing breaking changes. This can be
                               useful for excluding assemblies that are known
                               to not regress on certain .NET Framework
                               versions due to breaking changes. Note that,
                               currently, this parameter only affects breaking
                               change analysis; not portability analysis.
  -s, --suppressBreakingChange=VALUE
                             Specifies a breaking change (by ID) to suppress
                               during breaking change analysis. Any breaking
                               changes with IDs specified for suppression will
                               not be reported.
  -h, -?, --help             Show help

====================================================================================================
ApiPort.exe listOutputFormats [options]

Retrieves a list of output formats that are available in which to receive reports

  -e, --endpoint=VALUE       Service endpoint
  -h, -?, --help             Show help

====================================================================================================
ApiPort.exe listTargets [options]

Retrieves a list of targets available to analyze assemblies against

  -e, --endpoint=VALUE       Service endpoint
  -h, -?, --help             Show help
```

### DIFF
--- a/src/ApiPort/CommandLineOptions.cs
+++ b/src/ApiPort/CommandLineOptions.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using ApiPort.CommandLine;
+using ApiPort.Resources;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -13,8 +14,8 @@ namespace ApiPort
         private static IDictionary<string, Func<string, CommandLineOptionSet>> s_possibleCommands = new Dictionary<string, Func<string, CommandLineOptionSet>>(StringComparer.OrdinalIgnoreCase)
         {
             {"analyze", name => new AnalyzeOptionSet(name) },
-            {"listOutputFormats", name => new  ServiceEndpointOptionSet(name, AppCommands.ListOutputFormats) },
-            {"listTargets", name => new  ServiceEndpointOptionSet(name, AppCommands.ListTargets) },
+            {"listOutputFormats", name => new  ServiceEndpointOptionSet(name, AppCommands.ListOutputFormats, LocalizedStrings.CmdListOutputFormats) },
+            {"listTargets", name => new  ServiceEndpointOptionSet(name, AppCommands.ListTargets, LocalizedStrings.CmdListTargets) },
         };
 
         public static ICommandLineOptions ParseCommandLineOptions(string[] args)
@@ -47,11 +48,11 @@ namespace ApiPort
                 Console.WriteLine();
             }
 
-            Console.WriteLine("Available Commands:");
-
             foreach (var command in s_possibleCommands)
             {
-                Console.WriteLine($"- {command.Key}");
+                Console.WriteLine();
+                Console.WriteLine(new string('=', Math.Min(Console.WindowWidth, 100)));
+                command.Value(command.Key).Parse(new[] { "-?" });
             }
 
             return CommandLineOptionSet.ExitCommandLineOption;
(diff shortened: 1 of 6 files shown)


twsouthwick:
@mjrousos

src/ApiPort/CommandLineOptions.cs
```diff
@@ -47,11 +48,11 @@ private static ICommandLineOptions ShowHelp(string suppliedCommand = null)
                 Console.WriteLine();
             }
 
-            Console.WriteLine("Available Commands:");
-
             foreach (var command in s_possibleCommands)
             {
-                Console.WriteLine($"- {command.Key}");
+                Console.WriteLine();
+                Console.WriteLine(new string('=', Math.Min(Console.WindowWidth, 100)));
```
mjrousos: Nitpick on this; do we need to subtract one from Console.WindowWidth?

src/ApiPort/Resources/LocalizedStrings.resx
```diff
@@ -231,4 +232,13 @@ To learn more about how this tool works, including the data we are collecting, g
   <data name="CmdHelpSuppressBreakingChange" xml:space="preserve">
     <value>Specifies a breaking change (by ID) to suppress during breaking change analysis. Any breaking changes with IDs specified for suppression will not be reported.</value>
   </data>
+  <data name="CmdAnalyzeHelp" xml:space="preserve">
+    <value>Analyzes specified files/directories for IL and determines the APIs that are used. This information is analyzed for portability concerns or known breaking changes (depending on input parameters) and returns a report that gives a summary of issues and possible steps forward.</value>
```
mjrousos: This line is long; does it wrap funny? Should we maybe split it up at <= 80 characters per line?

mjrousos:
I added a couple nit-picky comments. Otherwise, it looks good.

src/ApiPort/CommandLineOptions.cs
twsouthwick: I never saw any issue with this. Have you seen a need for that in the past?

twsouthwick: Just played around with it and this seems to work fine.

mjrousos: If you have WindowWidth characters, won't the \r\n from WriteLine wrap to the next line so that if WindowWidth is >100, we end up with the next line directly below this one and with WindowWidth <= 100, we end up with a blank line in between?

Like I said, this is very nitpicky :)

mjrousos: If you tried it out and it looks fine, that's good enough for me. Thanks for checking!

PR #149, opened 2015-09-17
Use Microsoft.Framework.Configuration instead of NDesk.Options

twsouthwick:


src/ApiPort/CommandLine/ApiPortConfiguration.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            return options;
+        }
+
+        private static IEnumerable<ConfigMapCount> GetConfigMapCounts(Type type, Func<Type, bool> comparer, IDictionary<string, string> switchMapping)
+        {
+            var arrayNames = type.GetProperties(BindingFlags.Instance | BindingFlags.Public)
+                .Where(b => comparer(b.PropertyType))
+                .Select(prop => prop.Name)
+                .ToList();
```
conniey: nit: remove .ToList() so the evaluation of arrayNames happens lazily (while it is iterating through the foreach loop) instead of blocking here.

conniey:
Wow, great job Taylor. It looks good. :)

PR #158, opened 2015-10-20
Add breaking change for TreeViewItem

twsouthwick:


docs/BreakingChanges/141 WPF TreeViewItem must be used within a TreeView.md
```diff
@@ -0,0 +1,27 @@
...
+4.5
+
+### Source Analyzer Status
+Planned
+
+### Change Description
+Before 4.5, a TreeViewItem could be used outside of a TreeView. A change in 4.5 added a check during scrolling that assumes TreeViewItems is only used in the context of 
+TreeView. An example of when this is hit is if a TreeViewItem is the root of a DataTemplate or is used without being wihtin a TreeView.
+
+Description goes here.
```
mjrousos: The "Description goes here." text is no longer needed.

docs/BreakingChanges/141 WPF TreeViewItem must be used within a TreeView.md
```diff
@@ -0,0 +1,27 @@
+## ID: WPF TreeViewItem must be used within a TreeView
+
+### Scope
+Major
```
mjrousos: I would probably call this minor because ApiPort will only be able to detect TreeViewItem usage, and most of those will be within TreeViews. So, I think major is too strong of a scope.

docs/BreakingChanges/141 WPF TreeViewItem must be used within a TreeView.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Version Introduced
+4.5
+
+### Source Analyzer Status
+Planned
+
+### Change Description
+Before 4.5, a TreeViewItem could be used outside of a TreeView. A change in 4.5 added a check during scrolling that assumes TreeViewItems is only used in the context of 
+TreeView. An example of when this is hit is if a TreeViewItem is the root of a DataTemplate or is used without being wihtin a TreeView.
```
mjrousos: I think the issue only shows up if the TreeViewItem outside of a TreeView is also within a VirtualizingStackPanel. Sam would be the best person to give all those sorts of particulars.

mjrousos: And because customers probably aren't familiar with VirtualizingStackPanels, we will likely want to explain that those are any 'list'-type controls (ListBoxes, DataGrids, ListViews, etc.) which have virtualization enabled.

docs/BreakingChanges/141 WPF TreeViewItem must be used within a TreeView.md
```diff
@@ -0,0 +1,27 @@
...
+- [ ] Quirked
+- [ ] Build-time break
+
+### Recommended Action
+The TreeViewItem should be used from within a TreeView, so the XAML must be reworked to ensure that condition is met.
+
+### Affected APIs
+* Not detectable via API analysis
+
+[More information](LinkForMoreInformation)
```
conniey: Should there be a link for more information? Or is this just from the template? :)

docs/BreakingChanges/141 WPF TreeViewItem must be used within a TreeView.md
SamBent: line 14 'wihtin' -> 'within'.

The bug only shows up when
a. TreeViewItem's visual parent is not a panel.   (A TVI generated for a TreeView will have a panel as its parent)
b. The TVI is a descendant of a VirtualizingStackPanel acting as the "items host" for a list control (ListBox, DataGrid, ListView, etc.).   Virtualization doesn't need to be enabled.
c. The VSP is item-scrolling (ScrollUnit="Item").
d. Someone calls VSP.MakeVisible(v) to scroll an element v into view.   This can be done explicitly, or implicitly in a number of ways;  perhaps the most common way is simply clicking on v to give it the keyboard focus.
e. The visual-parent chain from v to the VSP passes through the TVI.

In other words, not only do you need to use a TreeViewItem outside of a TreeView, you also need the user to click on a descendant of the TVI to bring it into view.   If the TVI has no focusable descendants, you'll never see the bug.

docs/BreakingChanges/141 WPF TreeViewItem must be used within a TreeView.md
SamBent: I agree.   See my other comment about the full preconditions to see the bug - they're quite narrow.

Also, if there's some way to indicate to the public that it's just a bug, fixed in 4.6.2 (and in hotfix for 4.5.x, 4.6, 4.6.1).   Not an intentional breaking change.

mjrousos: We can give a 'version fixed' property which indicates the break is fixed in a given version (4.6.2, in this case). And we can call out the hotfix availability as a possible mitigation in 'recommended changes'

docs/BreakingChanges/141 WPF TreeViewItem must be used within a TreeView.md
twsouthwick: Whoops - removing...

docs/BreakingChanges/141 WPF TreeViewItem must be used within a TreeView.md
```diff
@@ -0,0 +1,32 @@
...
+Minor
+
+### Version Introduced
+4.5
+
+### Source Analyzer Status
+Planned
+
+### Change Description
+A change was introduced in 4.5 that restricts usage of TreViewItem elements outside of a TreeView. This manifests under the following conditions:
```
mjrousos: Typo: TreViewItem

mjrousos:
LGTM

Issue #130, opened 2015-08-31
Details/suggestions needed for some breaking changes


mjrousos:
Some breaking change files are missing details/suggestions:

* Microsoft/dotnet-apiport/docs/BreakingChanges/012 ComposablePartCatalog and its derived classes.md 
* Microsoft/dotnet-apiport/docs/BreakingChanges/015 Delegates pointing to extension methods lead to InvalidProgramException.md 
* Microsoft/dotnet-apiport/docs/BreakingChanges/016 Calling base method from an anonymous method causes exception.md

PR #161, opened 2015-10-23
Remove TODO items in the breaking change markdown files.

JJVertical:


mjrousos:
LGTM

twsouthwick:
Why did you remove some of the md files?

mjrousos:
Those were both breaking changes that came from customer reports (not MSDN) that we weren't able to repro (remember when someone - Charles? Connie?) was trying to write analyzers and hit a couple issues in a row that didn't seem to repro? @JJVertical has opened a couple internal issues against me to track closing on whether or not there's anything worth actually tracking in an md file for these.

twsouthwick:
Good to know

twsouthwick:
Fixes #130

PR #162, opened 2015-10-26
Changes to support only showing runtime issues

JJVertical:
This is the code changes. I will submit tests before completing the final pull request.

### DIFF
--- a/src/ApiPort/CommandLine/AnalyzeOptions.cs
+++ b/src/ApiPort/CommandLine/AnalyzeOptions.cs
@@ -30,6 +30,7 @@ namespace ApiPort.CommandLine
                 { "-r", "resultFormat" },
                 { "-p", "showNonPortableApis" },
                 { "-b", "showBreakingChanges" },
+                { "-u", "showRetargettingIssues" },
                 { "-i", "ignoreAssemblyFile" },
                 { "-s", "suppressBreakingChange" },
             };
@@ -49,6 +50,7 @@ namespace ApiPort.CommandLine
             public List<string> ResultFormat { get; set; }
             public bool ShowNonPortableApis { get; set; }
             public bool ShowBreakingChanges { get; set; }
+            public bool ShowRetargettingIssues { get; set; }
             public bool NoDefaultIgnoreFile { get; set; }
             public string IgnoreAssemblyFile { get; set; }
             public List<string> SuppressBreakingChange { get; set; }
@@ -114,6 +116,12 @@ namespace ApiPort.CommandLine
             {
                 if (options.ShowBreakingChanges)
                 {
+                    RequestFlags |= AnalyzeRequestFlags.ShowBreakingChanges;
+                }
+
+                if (options.ShowRetargettingIssues)
+                {
+                    RequestFlags |= AnalyzeRequestFlags.ShowRetargettingIssues;
                     RequestFlags |= AnalyzeRequestFlags.ShowBreakingChanges;
                 }
 
--- a/src/ApiPort/Resources/LocalizedStrings.resx
+++ b/src/ApiPort/Resources/LocalizedStrings.resx
@@ -239,33 +239,34 @@ go here - http://go.microsoft.com/fwlink/?LinkId=397652</value>
 for portability concerns or known breaking changes (depending on input parameters) and returns a report that 
 gives a summary of issues and possible steps forward.
 
-  -e, --endpoint=VALUE       Service endpoint
-  -f, --file=VALUE           [Required] Path to assembly file or directory of
-                               assemblies.
-  -o, --out=VALUE            Output file name
-  -d, --description=VALUE    Description of the submission
-  -t, --target=VALUE         The target you want to check against.
-  -r, --resultFormat=VALUE   The report output format
-  -p, --showNonPortableApis  Calculate non-portable APIs
-  -b, --showBreakingChanges  Calculate breaking changes on full .NET Framework
-      --noDefaultIgnoreFile  Do not use the standard assembly ignore list
-                               when analyzing breaking changes. The default
-                               ignore list can be found at KnownSafeBreaks.json
+  -e, --endpoint=VALUE         Service endpoint
+  -f, --file=VALUE             [Required] Path to assembly file or directory of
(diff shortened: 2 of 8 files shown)


src/Microsoft.Fx.Portability/Analysis/AnalysisEngine.cs
```diff
@@ -58,6 +62,14 @@ public IEnumerable<BreakingChangeDependency> FindBreakingChanges(IEnumerable<Fra
                     var breakingChanges = _recommendations.GetBreakingChanges(kvp.Key.MemberDocId).Distinct();
                     foreach (var b in breakingChanges)
                     {
+                        if (onlyShowRuntimeIssues == true)
```
mjrousos: The '== true' isn't actually necessary here.

src/ApiPort/Resources/LocalizedStrings.resx
```diff
@@ -248,6 +248,11 @@ gives a summary of issues and possible steps forward.
   -r, --resultFormat=VALUE   The report output format
   -p, --showNonPortableApis  Calculate non-portable APIs
   -b, --showBreakingChanges  Calculate breaking changes on full .NET Framework
+  -u, --onlyShowRuntimeIssues
+                             Only show the issues that would need to be 
+                             addressed if keeping existing target and 
+                             running on a newer framework version. This 
+                             will automatically show breaking changes.
```
mjrousos: This help statement says that --u will imply --b, but it doesn't look like that's the case. You should either update argument parsing to make sure that happens, or update the definition of the OnlyShowRuntimeChanges flag to include the ShowBreakingChangesFlag.

src/Microsoft.Fx.Portability/Analysis/AnalysisEngine.cs
```diff
@@ -58,6 +62,11 @@ public IEnumerable<BreakingChangeDependency> FindBreakingChanges(IEnumerable<Fra
                     var breakingChanges = _recommendations.GetBreakingChanges(kvp.Key.MemberDocId).Distinct();
                     foreach (var b in breakingChanges)
                     {
+                        if (showRetargettingIssues == false && b.IsRetargeting)
```
mjrousos: Could be simplified to !showRetargetingIssues && b.IsRetargeting

mjrousos:
LGTM

src/ApiPort/CommandLine/AnalyzeOptions.cs
```diff
@@ -30,6 +30,7 @@ public override ICommandLineOptions Parse(IEnumerable<string> args)
                 { "-r", "resultFormat" },
                 { "-p", "showNonPortableApis" },
                 { "-b", "showBreakingChanges" },
+                { "-u", "ShowRetargettingIssues" },
```
twsouthwick: can you change the casing to be consistent with the other options?

src/ApiPort/CommandLine/AnalyzeOptions.cs
```diff
@@ -112,6 +114,11 @@ public override IEnumerable<string> InvalidInputFiles
 
             private void UpdateRequestFlags(Options options)
             {
+                if (options.ShowRetargettingIssues)
+                {
+                    RequestFlags |= AnalyzeRequestFlags.ShowRetargettingIssues;
```
twsouthwick: Do we also want it to imply ShowBreakingChanges?

src/ApiPort/Resources/LocalizedStrings.resx
```diff
@@ -248,6 +248,7 @@ gives a summary of issues and possible steps forward.
   -r, --resultFormat=VALUE   The report output format
   -p, --showNonPortableApis  Calculate non-portable APIs
   -b, --showBreakingChanges  Calculate breaking changes on full .NET Framework
+  -u, --ShowRetargettingIssues Include the retargetting issues in the reports
```
twsouthwick: Again, casing

src/ApiPort/CommandLine/AnalyzeOptions.cs
mjrousos: Good point. --u doesn't mean anything with --b; it probably ought to enable ShowBreakingChanges automatically.

src/ApiPort/Resources/LocalizedStrings.resx
```diff
@@ -248,6 +248,7 @@ gives a summary of issues and possible steps forward.
   -r, --resultFormat=VALUE   The report output format
   -p, --showNonPortableApis  Calculate non-portable APIs
   -b, --showBreakingChanges  Calculate breaking changes on full .NET Framework
+  -u, --showRetargettingIssues Include the retargetting issues in the reports
```
twsouthwick: I just noticed this offsets the whitespace. Can you adjust the other lines so they look more in line?

src/ApiPort/CommandLine/AnalyzeOptions.cs
```diff
@@ -117,6 +119,17 @@ private void UpdateRequestFlags(Options options)
                     RequestFlags |= AnalyzeRequestFlags.ShowBreakingChanges;
                 }
 
+                if (options.ShowRetargettingIssues)
+                {
+                    RequestFlags |= AnalyzeRequestFlags.ShowRetargettingIssues;
+
+                    //ShowRetargettingIssues doesn't mean much without ShowBreakingChanges turned on
+                    if (!RequestFlags.HasFlag(AnalyzeRequestFlags.ShowBreakingChanges))
```
mjrousos: This will work, but for what it's worth |= the same thing multiple times is a no-op, so you could just add the ShowBreakingChanges flag unconditionally here.

JJVertical: I thought about doing it this way, but thought the logic of the code would be more clear this way since I am already setting the ShowBreakingChanges in the if statement above.

twsouthwick: I agree with Mike - the extra check isn't needed and makes the logic harder to follow.

twsouthwick:
This fixes  #118

PR #164, opened 2015-10-27
Clean up titles and ensure file name and title match for issues up to 040

JJVertical:


docs/BreakingChanges/018 BlockingCollectionT.TryTakeFromAny does not throw anymore.md
```diff
@@ -1,4 +1,4 @@
-## 18: BlockingCollection<T>.TryTakeFromAny doesn't throw anymore
+## 18: BlockingCollection<T>.TryTakeFromAny does not throw anymore
```
mjrousos: While we're fixing this one up, may want to do something about the \<T\> not showing up in the title. Maybe just escape the <> with \\.

docs/BreakingChanges/005 ListT.ForEach can throw exception when modifying list item.md
```diff
@@ -1,4 +1,4 @@
-## 5: `List<T>.ForEach`
+## 5: ListT.ForEach can throw exception when modifying list item
```
mjrousos: ListT instead of List\<T\>? Unless you think the \\ to escape is too ugly in the raw mode, it might be nice to use \<T\> since it's more familiar syntax.

docs/BreakingChanges/013 Missing Target Framework Moniker results in 4.0 behavior.md
```diff
@@ -1,4 +1,4 @@
-## 13: Missing TFM results in 4.0 behavior
+## 13: Missing Target Framework Moniker results in 4.0 behavior
```
mjrousos: Thanks for fixing this! The fact that this file's name was the exact opposite of the actual behavior has been bugging me for some time!

PR #168, opened 2015-10-29
Updating help instructions and targets

conniey:
Updating Introduction with the latest command-line options and targets that we support.


twsouthwick:
LGTM

PR #170, opened 2015-10-29
Fixing line endings on BreakingChanges.md

conniey:
This fixes the line endings on BreakingChanges.md that was missed


twsouthwick:
Is this still needed? I'm not seeing any line endings that are wrong on master now

conniey:
Weird. if I do a `git clone https://github.com/Microsoft/dotnet-apiport` then `cd` into the directory, I see that file.  What are your settings for `git config --global -e`?

```
[core]
        editor = vim
        autocrlf = true
        safecrlf = true
        preloadindex = true
        fscache = true
```

twsouthwick:
I see it now.

PR #188, opened 2015-11-10
Adding logging and error checking to the report writer

conniey:
The report writer can report false positives. (ie. That a report was successfully written when it was not.)

This adds checks to ensure that the proper status is shown to users when a report cannot be written. It modifies the output of `ApiPortClient.WriteAnalysisReportsAsync` to only contain the paths of reports that were successfully written to.


mjrousos:
LGTM

twsouthwick:
LGTM

PR #191, opened 2015-11-14
Add runtimes : { win : {} } to project.json files

mjrousos:
This is necessary to meet VS2015 Update 1 tools requirements


src/ApiPort/project.json
```diff
@@ -1,6 +1,7 @@
 ﻿{
   "runtimes": {
-    "win-anycpu": { }
+    "win-anycpu": { },
+    "win": { }
```
twsouthwick: Do we need both of these?

twsouthwick:
LGTM - should we wait until RTM of Update 1? In either case, please update the README.md to clarify that Update 1 is now required.

src/ApiPort/project.json
mjrousos: I left the old one because Microsoft/msbuild#344 mentioned that the win- style is needed to build with RTM.

mjrousos:
Ah, does this break building with RTM? If so, we should wait for Update 1 to release (RTM is such an overloaded term here :). I will test it out; I thought by not removing the old win- runtimes, it would still build without update 1 but I need to dig up a non-upgraded machine to try it.

src/ApiPort/project.json
mjrousos: I checked and it looks like win-anycpu is still needed to build on 2015 RTM

mjrousos:
I tried it out and these projects seem to build fine on both RTM and Update 1 once the project.json files are updated.

PR #198, opened 2015-11-18
Apply heuristic to only collect APIs that are most likely .NET framework APIs

### LINKED ISSUE #185: Add a heuristic to only include framework APIs
The tool currently gathers all APIs. We could come up with a simple heuristic that takes into account the following:

- Known strong name keys since all framework assemblies are signed
- Assembly starts with System, Microsoft, mscorlib, (maybe Mono?)
- Namespace starts with System, Microsoft, (maybe Mono?)

twsouthwick:
This fixes #185 by adding a filter into the dependency finder that attempts to determine whether or not a given API is a framework API by looking at the assembly it is defined in. If that assembly:

1. Starts with Microsoft, System, or Mono
2. Is signed with a known Microsoft public key token
3. Is mscorlib

We will include them in the request. Otherwise, the APIs are discarded.

@terrajobst  @mjrousos


mjrousos:
At one point way back when, I compiled [this](https://sources.visualstudio.com/DefaultCollection/NetEcosystem/_git/ApiSequencer#path=%2FApiSequencerCmd%2FNetFXAssemblies.txt&version=GBmaster&_a=contents) list of NetFX/CLR assemblies. It's imperfect (I think there are some native things on that list and it's likely missing some other items), so don't just take it wholesale, but some observations from it:

* Presentation* may be interesting for WPF things
* Similarly, UIAutomation* may be interesting for WPF UIA
* These may be caught by PKT checks, but some other non-system and non-Microsoft NetFX assemblies are: 
  * WindowsBase.dll
  * ReachFramework.dll
  * WindowsFormsIntegration.dll
  * Accessibility.dll
  * CustomMarshallers.dll
  * SysGlobal.dll
  * ISymWrapper.dll
  * XamlBuildTask.dll
  * Mscorlib.Extensions.dll (this is in some CoreCLR profiles and shouldn't be built against directly, but it may not hurt to include it for completeness)

src/Microsoft.Fx.Portability.MetadataReader/AssemblyReferenceInformation.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public Version Version { get; }
+
+        public string PublicKeyToken { get; }
+
+        public override string ToString() => _string;
+
+        public override bool Equals(object obj)
+        {
+            return string.Equals(_string, obj.ToString(), StringComparison.Ordinal);
```
mjrousos: Not OrdinalIgnoreCase?

src/Microsoft.Fx.Portability.MetadataReader/DotNetFrameworkFilter.cs
```diff
@@ -0,0 +1,56 @@
...
+            {
+                return true;
+            }
+
+            if (_frameworkAssemblyNamePrefixes.Any(p => assembly.Name.StartsWith(p, StringComparison.OrdinalIgnoreCase)))
+            {
+                return true;
+            }
+
+            if (string.Equals(assembly.Name, "mscorlib"))
```
mjrousos: StringComparison.OrdinalIgnoreCase?

src/ApiPort/FileOutputApiPortService.cs
```diff
@@ -87,7 +87,7 @@ private void WriteOutput(AnalyzeRequest a)
                 Targets = new SortedSet<string>(a.Targets, StringComparer.Ordinal),
                 ApplicationName = a.ApplicationName,
                 Version = a.Version,
-                BreakingChangesToSuppress = new SortedSet<string>(a.BreakingChangesToSuppress, StringComparer.Ordinal),
+                BreakingChangesToSuppress = new SortedSet<string>(a.BreakingChangesToSuppress ?? Enumerable.Empty<string>(), StringComparer.Ordinal),
```
mjrousos: Thanks for fixing this!

src/Microsoft.Fx.Portability.MetadataReader/DotNetFrameworkFilter.cs
```diff
@@ -0,0 +1,59 @@
...
+            "b77a5c561934e089", // ECMA
+            "b03f5f7f11d50a3a", // DEVDIV
+            "7cec85d7bea7798e", // SLPLAT
+            "31bf3856ad364e35", // Windows
+            "24eec0d8c86cda1e", // Phone
+            "0738eb9f132ed756", // Mono
+            "ddd0da4d3e678217", // Component model
+            "84e04ff9cfb79065", // Mono Android
+            "842cf8be1de50553"  // Xamarin.iOS
+        }, StringComparer.OrdinalIgnoreCase);
```
twsouthwick: This is a list of all strong names in the %ProgramFiles%\Reference Assemblies\Microsoft directories. It catches all of the assemblies you listed.

src/Microsoft.Fx.Portability.MetadataReader/AssemblyReferenceInformation.cs
```diff
@@ -0,0 +1,39 @@
...
+        public override string ToString() => _string;
+
+        public override bool Equals(object obj)
+        {
+            return string.Equals(_string, obj.ToString(), StringComparison.OrdinalIgnoreCase);
+        }
+
+        public override int GetHashCode()
+        {
+            return _string.GetHashCode();
```
mjrousos: Should have mentioned this last time around, but if we allow assembly equality to be case-insensitive (which I think makes sense), then you should probably ToLower the _string here to make sure hash codes are case-sensitive.

twsouthwick: I'm using the StringCOmparison.OrdinalIgnoreCase for the equality and gethashcode now. The file has moved (I moved it to the core library instead of the metadata reader)

mjrousos:
LGTM

PR #200, opened 2015-11-19
Add an argument check in ApiPortService

twsouthwick:
While updating the website, a null got passed in as the endpoint, it was tricky to figure out what was happening. This will instead verify it is not a null or whitespace, and then set the 

@JJVertical


conniey:
LGTM

PR #201, opened 2015-11-20
Abstract file handling to IAssemblyFile

twsouthwick:
@mayaganesan started some work over the summer to implement file parsing using byte[]. Instead of byte[], it would be better to abstract what we need from an input file, so this creates an IAssemblyFile interface that gives a path, status of whether or not it exists, and can open a stream for reading.


src/ApiPort/CommandLine/AnalyzeOptions.cs
```diff
@@ -191,10 +192,23 @@ public int Compare(FileInfo x, FileInfo y)
                     }
 
 
-                    return x.FullName.CompareTo(y?.FullName);
+                    return x.Path.CompareTo(y?.Path);
```
mjrousos: It seems like we should rely on the IAssemblyFile implementation to define the comparer. Comparing paths works for FilePathAssemblyFile, but I can imagine IAssemblyFile implementations that represent assemblies in memory or something like that. In those cases, paths might always be null and there would have to be some other mechanism for comparison (perhaps that IAssemblyFile implementation would define an assembly name field and use that, for example).

src/Microsoft.Fx.Portability/IAssemblyFile.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using System.IO;
+
+namespace Microsoft.Fx.Portability
+{
+    public interface IAssemblyFile
+    {
+        Stream OpenRead();
+
+        string Path { get; }
```
mjrousos: Why path instead of full name? It seems like a name would be a more universally available. I think path is needed for a lot of our current MetadataReader-based code, but it we really want to support more general IAssembly-style inputs, we may need to allow for in-memory assemblies that report their own identities, etc.

twsouthwick: Good idea. I'll switch Path -> FullName

src/ApiPort/CommandLine/AnalyzeOptions.cs
twsouthwick: This comparer is specifically for use in ApiPort so that we sort according to display name. By switch to FullName instead of Path, then in memory assemblies will work. I'll have to check how it handles a null FullName, but I'd prefer for it fail in that case (ie, FullName should always be defined)

src/Microsoft.Fx.Portability/IAssemblyFile.cs
mjrousos: Seems good (in that it's a more general interface). Do we know if this will work with our metadata reader? A quick perusal shows that we attempt to gather assembly information by loading from disk. Are more changes needed there to enable a non-file-based assembly to work?

twsouthwick: Good question - I'll add a test using a byte[] wrapper

mjrousos: Cool. The places I'm thinking of that could be problematic are, for example, `MetadataReaderExtensions.GetAssemblyInfo` [here](https://github.com/Microsoft/dotnet-apiport/blob/master/src/Microsoft.Fx.Portability.MetadataReader/MetadataReaderExtensions.cs)

twsouthwick: :) I actually fixed this so that IAssemblyFile contains a Version property.

twsouthwick: Added a test, and it seems to work fine.

tests/Microsoft.Fx.Portability.MetadataReader.Tests/ManagedMetadataReaderTests.cs
```diff
@@ -245,5 +249,25 @@ public bool IsFrameworkAssembly(AssemblyReferenceInformation assembly)
...
+
+            public StreamFromFileAssemblyFile(string path)
+            {
+                _path = path;
+                _bytes = File.ReadAllBytes(path);
+            }
+
+            public bool Exists { get; } = true;
+
+            public string FullName => _path;
```
mjrousos: Hm... this still isn't exactly what I had in mind because it still has a path.

If a goal of this interface is to allow assemblies to be loaded from byte[]s, then we should consider what to do if the user has an assembly in memory that doesn't exist on disk. That's why I thought requiring a path property might not make sense (all assemblies have some identity - a 'name' - but not all assemblies necessarily have a path).

I suspect, though, that we still need a path for our MetadataReader interactions. If that's the case, let's just keep calling this API 'path'. But, we should think about another work item down the road to enable loading an assembly that came from some other location (web service call? reflection.emit? compiled by Roslyn but not emitted?).

twsouthwick: The FullName can be anything - it's required for the analysis to know what the "name" is. We can change it to Path, but conceptually, it's just the name of the input assembly so the reports have something to identify it with. We can call it "Name" since FullName doesn't really mean anything outside of the Path usage

mjrousos: I like 'Name' as the property name here; I agree that covers all the scenarios well.

Have we tested any scenarios where the name isn't a path?

twsouthwick: Yes. I've updated the tests to create various implementations of IAssemblyFile: 

- ResourceStreamAssemblyFile will take a resource name and stream the file from there
- CSharpCompileAssemblyFile will compile an assembly and emit the result to a byte[] that is wrapped in a stream

Neither of these set the Name (there is a test that does set the Name, though). The only places Name is used is to list assemblies it fails to open or non existent files.

mjrousos: Great; thanks!

mjrousos:
LGTM

PR #207, opened 2015-11-25
Copy license terms to ApiPort build

conniey:
Copy license terms to ApiPort build output.
Miscellaneous changes:
- Fix nuget license URI
- Remove deleted build files from solution


mjrousos:
LGTM

PR #208, opened 2015-11-30
Importing recommended changes from catalog

conniey:
This imports the existing recommended changes to be consumed by PortabilityService

@twsouthwick, @terrajobst

### DIFF
--- a/docs/RecommendedChanges/! Template.md
+++ b/docs/RecommendedChanges/! Template.md
@@ -1,0 +1,5 @@
+### Recommended Action
+The action to perform for the following APIs
+
+### Affected APIs
+* `T:System.Type`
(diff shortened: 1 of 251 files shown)


twsouthwick:
Awesome! It feels like having a folder hierarchy all the way to the member/property is a lot. What about having it at the containing namespace level? ie recommended changes for all things in System.Net would be in System/Net/[recommended changes]. This would reduce some of the duplication of recommended changes.

conniey:
@twsouthwick That makes sense! I just updated the PR with changes belonging to the same namespace... Does this look better. :)

twsouthwick:
Much better! LGTM

twsouthwick:
btw - why no spaces in the file names?

conniey:
I had manually created names for them that way. I fixed them to include spaces. :)

PR #210, opened 2015-12-01
ApiPort for .NET Core (including x-plat console host)

### LINKED ISSUE #117: Switch ApiPort to use DNX
Currently ApiPort targets .NET Framework 4.6, but we could switch it to use DNX on .NET Core so it doesn't have the 4.6 requirement.

mjrousos:
This adds a CoreCLR ApiPort build. The changes to ApiPort itself are fairly straightforward, though the magic of a csproj that builds as a .NET Core exe was more complicated.

A few things to know about this PR -
* This upgrades System.Reflection.Metadata (because the version we were using previously did not support .NET Core). I had to use a very new alpha version to get one that both targeted .NET Core and had all the features we need. 
* With these changes, it's necessary to disable NuGet restore in VS (as per [these](http://aka.ms/port2core) instructions) or there will be some false errors (the error list shows issues, but the build completes successfully). NuGet restore can be left on if you don't mind noise, but otherwise restoring from the command line works cleanly.
* Because the CoreCLR ApiPort host is OS- and architecture- specific, there are a lot more configurations in the sln file now. For everything except ApiPort.Core, the projects build the same for all of them. That project needs to target each individually, though, in order to get the correct hosting bits.
* This has been tested on Linux (Ubuntu 14.04) but is not yet tested on Mac (my MacBook is in the mail on its way to me currently). There may be a few Mac-specific tweaks necessary.
* Because System.Configuration isn't available for .NET Core, Unity configuration via config file is only enabled for Desktop ApiPort. We should look into run-time configuration through alternate means for the future. I think it's ok to merge this as is in the short-term, though. It just means that offline-mode is a to-be-implemented feature for ApiPort.Core at this point.

Fixes #117

@twsouthwick @conniey


twsouthwick:
Would it be better to have an ApiPort.Core solution that's separate and pulls in the nuget packages it needs?

twsouthwick:
As far as Unity stuff is concerned, we could switch it to use Autofac, which has a nice, modern configuration system available

src/ApiPort/ApiPort.csproj
```diff
@@ -13,6 +13,7 @@
     <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+    <DefineConstants>DESKTOP_CONFIGURATION;RICH_CONSOLE</DefineConstants>
```
twsouthwick: The rest of the solution uses the convention FEATURE_* for define constants (such as FEATURE_DESKTOP_CONFIGURATION and FEATURE_RICH_CONSOLE). Can you switch these to follow that?

src/ApiPort/ApiPort.project.json
```diff
@@ -8,8 +8,8 @@
     "Microsoft.Framework.Configuration.CommandLine": "1.0.0-beta5",
     "Microsoft.Framework.OptionsModel": "1.0.0-beta5",
     "Newtonsoft.Json": "6.0.8",
-    "System.Collections.Immutable": "1.1.37",
-    "System.Reflection.Metadata": "1.1.0-alpha-00008",
+    "System.Collections.Immutable": "1.1.38-*",
+    "System.Reflection.Metadata": "1.2.0-alpha-00017",
```
twsouthwick: :) @conniey  was looking into updating this but there was a bug in the metadata reader she found that was fixed in this release.

src/ApiPort/CommandLineOptions.cs
```diff
@@ -70,9 +70,17 @@ private static ICommandLineOptions ShowHelp(string suppliedCommand = null, bool
                 // TODO: Get invalid parameter (Microsoft.Framework.Configuration currently does not surface this)
                 Program.WriteColorLine($"Invalid parameter passed to {suppliedCommand}", ConsoleColor.Red);
             }
-
-            var codebase = new Uri(typeof(CommandLineOptions).GetTypeInfo().Assembly.CodeBase);
-            var path = Path.GetFileName(codebase.AbsolutePath);
+            
+            var location = typeof(CommandLineOptions).GetTypeInfo().Assembly.Location;
+            var path =
+#if NETCORE
```
twsouthwick: Can you use FEATURE_ASSEMBLY_LOCATION? I think this is already defined (maybe in a different project, though)

src/ApiPort/ConsoleProgressReporter.cs
```diff
@@ -77,6 +82,8 @@ public ConsoleProgressTask(string task, int? totalCount)
             {
                 await Task.Delay(1);
 
+#if RICH_CONSOLE
```
twsouthwick: Should we create a separate ConsoleProgressREporter? Will this still show up in the platforms that support it?

src/ApiPort/ApiPort.project.json
twsouthwick: Yup - it was a corner case causing a failure on methods with an arglist parameter. It should be fixed in alpha-17 - are you seeing any tests failing? There was one test failing, so if all tests are passing, it's fixed.

mjrousos: Yep, all tests are passing with this one.

src/ApiPort/CommandLineOptions.cs
mjrousos: Sure; I'll switch to that.

src/ApiPort/CommandLine/AnalyzeOptions.cs
```diff
@@ -210,7 +210,21 @@ public FilePathAssemblyFile(string path)
...
+                {
+                    get
+                    {
+                        try
+                        {
+                            return FileVersionInfo.GetVersionInfo(_path).FileVersion;
+                        }
+                        catch (ArgumentException)
+                        {
+                            // Temporary workaround for CoreCLR-on-Linux bug (dotnet/corefx#4727) that prevents get_FileVersion from working on that platform
```
twsouthwick: Can you add that the fix should be available with RC2 of .NET Core?

src/ApiPort/ConsoleProgressReporter.cs
mjrousos: I don't think there's any need to make a separate ConsoleProgressReporter since this is already #Ifdef'd out so easily. As it is now, everything works as before on Windows and on Ubuntu/Mac we just don't have the progress reporter's waiting animation (but the tool still seems perfectly usable without it).

mjrousos:
We could put ApiPort.Core in a separate solution and get NuGet packages; I just put it in the existing solution to mirror how our other projects supported .NET Core/Desktop versions. Is there a reason you prefer it to be separated out?

And switching to Autofac seems good. I'll create an issue to take care of that (though not necessarily as part of this PR).

src/ApiPort/ConsoleProgressReporter.cs
twsouthwick: Sounds good

src/ApiPort/CommandLine/AnalyzeOptions.cs
mjrousos: Yes, good idea.

twsouthwick:
I suggested a separate project since this one introduces so many different configurations. I'm fine with the way it is, though

mjrousos:
Ah, yea, I can see wanting to keep the sln clean. I think it's probably ok as it is, though (I don't look at the sln that often anyhow). I'll leave it for now and we can extract it later, if needed (I did play around with pulling Microsoft.Fx.Portability from NuGet early on and it all works fine, as expected).

src/ApiPort/CommandLineOptions.cs
mjrousos: Actually, I just looked at this one and don't think that FEATURE_ASSEMBLY_LOCATION is right. I remember, now, why I used the general 'NETCORE' define instead of something more specific.

The difference here is whether or not we include the current file's extension in the example command lines in the help messages. I want to drop the extension in the CoreCLR case, since the extension will be .dll and that's not the file that users should be calling. In hosted cases, the user should actually call ApiPort or ApiPort.exe (depending on the platform), so I just leave the extension off.

twsouthwick: Got it. That makes sense

twsouthwick:
Can you also bump the Microsoft.Fx.Portability.MetadataReader library version since we're increasing the System.Reflection.Metadata reference? Other than that, LGTM

twsouthwick:
Can you update the offline mode section in the docs to mention it is not available on core at the moment?

PR #213, opened 2015-12-02
Add breaking change for BinaryFormatter issue

twsouthwick:


docs/BreakingChanges/147 BinaryFormatter can fail to find type from LoadFrom context.md
```diff
@@ -0,0 +1,44 @@
...
+Edge
+
+### Version Introduced
+4.5
+
+### Source Analyzer Status
+Planned
+
+### Change Description
+As of .NET Framework 4.5, a number of XmlSerializer changes may cause differences in deserialization when using `BinaryFormatter` to deserialize types that had been loaded in the
```
mjrousos: What's the connection between XmlSerializer and the BinaryFormatter? I thought they were separate serialization technologies.

docs/BreakingChanges/147 BinaryFormatter can fail to find type from LoadFrom context.md
```diff
@@ -0,0 +1,44 @@
...
+LoadFrom context. The default serialization binder does not automatically search the LoadFrom context, although it may have worked in some circumstances based on the old behavior
+of XmlSerializer. Due to the changes, a FileNotFoundException may be thrown that the Assembly it was defined in could not be found.
```
conniey: > Due to the changes, a FileNotFoundException may be thrown that the Assembly it was defined in could not be found.

I am not sure I understand this sentence. Does it mean sometimes a FileNotFoundException is thrown when it cannot find the assembly that the class was defined in?

docs/BreakingChanges/147 BinaryFormatter can fail to find type from LoadFrom context.md
twsouthwick: Does it make sense now? The only instance we've seen is when XmlSerializer is used in conjunction with BinarySerializer. Seems like things get loaded into a different context or with different timing than in earlier versions.

docs/BreakingChanges/147 BinaryFormatter can fail to find type from LoadFrom context.md
twsouthwick: Yeah, it's a bit awkward, but your understanding is right. I've updated the wording.

docs/BreakingChanges/147 BinaryFormatter can fail to find type from LoadFrom context.md
mjrousos: Yea, this is a little better. Do we have any more details about what exactly the XmlSerializer has to do to load an assembly in this weird way? If we do, that would be a good 'More Info' link. I think as it is I probably have enough information to make sense of the ApiPort result, but I wouldn't have enough understanding to go and write a good analyzer for this.

twsouthwick: I don't think we have any more information. I was thinking that an analyzer would look for BinaryFormatter, and create a custom Binder if one was not already set as an informational

mjrousos: I see. Yea, that may work for an analyzer. It's too bad we can't look for the XmlSerializer, though, because I expect that the majority of BinaryFormatter usages don't also use an XmlSerializer (they're often treated as 'competing' options for how to serialize something). We may want to at least scope the analyzer to look for some XmlSerializer usage in the project, even if we don't know exactly what pattern to look for.

mjrousos:
LGTM

docs/BreakingChanges/147 BinaryFormatter can fail to find type from LoadFrom context.md
twsouthwick: Good idea.

PR #215, opened 2015-12-02
Move the 'sending data to service' progress reporting to a higher level

mjrousos:
If multiple report writers are specified (via -r), GetResultFromServiceAsync may be run multiple times in parallel. Progress reporters don't necessarily give good/pretty output when run in this way, so move the progress report usage up-stack to avoid the in-parallel case.

***Note***   For some reason ApiPort.cs had an odd encoding that caused git to treat it as a binary file. I have two commits in this PR - the first is simply saving the file with a new encoding. You can see the actual code change by clicking on the second commit.


conniey:
Hmm.. I still see it as a binary instead of a text file. :(

BIN Binary file modified  src/Microsoft.Fx.Portability/ApiPortClient.cs  
Binary file not shown

mjrousos:
Yes, that because the original file was encoded bizarrely so the left side of the diff is always funky, even after I've fixed it. If you look at the individual commits in the PR, though, you should see that the second one displays correctly (since the first one is where I fixed the encoding issue).

Does that not work?

conniey:
When I checked it out locally, I could see the changes.  I wish I knew what was happening with all these encodings because I've seen it before.

LGTM.

PR #216, opened 2015-12-03
Improve logging for InternalServerErrors

conniey:
This improves the output when an error occurs in the PortabilityService.  This now shows the exception that occurs in the server.

@mjrousos, @twsouthwick

### DIFF
--- a/PortabilityTools.sln
+++ b/PortabilityTools.sln
@@ -40,6 +40,7 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Fx.Portability.Re
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "build", "build", "{FEEF35CB-3AC4-40C5-9CB0-85E52E38D5C6}"
 	ProjectSection(SolutionItems) = preProject
+		build\Get-CatalogFile.ps1 = build\Get-CatalogFile.ps1
 		build\Get-Nuget.ps1 = build\Get-Nuget.ps1
 		build\Microsoft.Fx.Portability.MetadataReader.nuspec = build\Microsoft.Fx.Portability.MetadataReader.nuspec
 		build\Microsoft.Fx.Portability.nuspec = build\Microsoft.Fx.Portability.nuspec
--- a/src/Microsoft.Fx.Portability/CompressedHttpClient.cs
+++ b/src/Microsoft.Fx.Portability/CompressedHttpClient.cs
@@ -1,10 +1,8 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using Microsoft.Fx.Portability.Reporting.ObjectModel;
 using Microsoft.Fx.Portability.Resources;
 using System;
-using System.Diagnostics;
 using System.Globalization;
 using System.Net;
 using System.Net.Http;
@@ -130,6 +128,15 @@ namespace Microsoft.Fx.Portability
                             throw new NotFoundException();
                         case HttpStatusCode.Unauthorized:
                             throw new UnauthorizedEndpointException();
+                        case HttpStatusCode.InternalServerError:
+                            var content = await response.Content?.ReadAsStringAsync();
+
+                            if (string.IsNullOrEmpty(content))
+                            {
+                                throw new PortabilityAnalyzerException(string.Format(CultureInfo.CurrentCulture, LocalizedStrings.UnknownInternalErrorCodeMessage, response.StatusCode, response.ReasonPhrase));
+                            }
+
+                            throw InternalServerErrorException.Create(response.ReasonPhrase, content);
                     }
 
                     throw new PortabilityAnalyzerException(string.Format(CultureInfo.CurrentCulture, LocalizedStrings.UnknownErrorCodeMessage, response.StatusCode));
--- a/src/Microsoft.Fx.Portability/InternalServerErrorException.cs
+++ b/src/Microsoft.Fx.Portability/InternalServerErrorException.cs
@@ -1,0 +1,22 @@
+﻿using Microsoft.Fx.Portability.Resources;
+using System;
+
+namespace Microsoft.Fx.Portability
+{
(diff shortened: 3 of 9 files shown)


mjrousos:
Several of the files don't show diffs because git thinks they're binary files. Can you check their encoding and update it to something git recognizes if needed?

PortabilityTools.sln
```diff
@@ -40,6 +40,7 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Fx.Portability.Re
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "build", "build", "{FEEF35CB-3AC4-40C5-9CB0-85E52E38D5C6}"
 	ProjectSection(SolutionItems) = preProject
+		build\Get-CatalogFile.ps1 = build\Get-CatalogFile.ps1
```
mjrousos: Hasn't this file existed for a while? Is the change just to make it appear in the solution explorer?

src/Microsoft.Fx.Portability/CompressedHttpClient.cs
```diff
@@ -3,6 +3,7 @@
 
 using Microsoft.Fx.Portability.Reporting.ObjectModel;
 using Microsoft.Fx.Portability.Resources;
+using Newtonsoft.Json;
```
mjrousos: Why is this needed? I may be overlooking it, but I don't see anything from the namespace used in the new code below.

src/Microsoft.Fx.Portability/CompressedHttpClient.cs
```diff
@@ -130,8 +131,19 @@ public CompressedHttpClient(ProductInformation info)
...
+
+                            if (string.IsNullOrEmpty(content))
+                            {
+                                throw new PortabilityAnalyzerException(string.Format(CultureInfo.CurrentCulture, LocalizedStrings.UnknownInternalErrorCodeMessage, response.StatusCode, response.ReasonPhrase));
+                            }
+
+                            throw InternalServerErrorException.Create(response.ReasonPhrase, content);
                     }
 
+                    var error = await response.Content?.ReadAsStringAsync();
```
mjrousos: It looks like the error local is unused.

src/Microsoft.Fx.Portability/Resources/LocalizedStrings.resx
```diff
@@ -331,4 +331,13 @@
   <data name="CouldNotWriteReport" xml:space="preserve">
     <value>Could not write report to directory '{0}' with file name '{1}' and extension '{2}'.</value>
   </data>
+  <data name="UnknownInternalErrorCodeMessage" xml:space="preserve">
+    <value>There was an issue retrieving error message from the service. Reason: '{0}'. Please wait while we update the service.</value>
```
mjrousos: Nit-pick: "There was an issue retrieving error message from the service" reads a bit awkwardly. It might be better as "retrieving **an** error message" or "retrieving **the** error message".

conniey:
> Never mind, all seems good now. If you fixed it, thanks! If not, that was weird.

CodeFormatter was saving the changed files in a format that git could not understand. I had to manually re-save them as UTF-8.

PortabilityTools.sln
conniey: Yes.

src/Microsoft.Fx.Portability/CompressedHttpClient.cs
conniey: I thought CodeFormatter would remove all the unused usings.. but it does not. I removed those.

src/Microsoft.Fx.Portability/CompressedHttpClient.cs
conniey: Fixed!

src/Microsoft.Fx.Portability/Resources/LocalizedStrings.resx
conniey: It does sound awkward. Fixed.

mjrousos:
LGTM

src/Microsoft.Fx.Portability/CompressedHttpClient.cs
```diff
@@ -130,6 +128,15 @@ public CompressedHttpClient(ProductInformation info)
                             throw new NotFoundException();
                         case HttpStatusCode.Unauthorized:
                             throw new UnauthorizedEndpointException();
+                        case HttpStatusCode.InternalServerError:
+                            var content = await response.Content?.ReadAsStringAsync();
```
twsouthwick: This can cause a null ref. If response.Content is null, you'll get 'await null' which will fail.

src/Microsoft.Fx.Portability/InternalServerErrorException.cs
```diff
@@ -0,0 +1,25 @@
...
+        private InternalServerErrorException(string message, string exception)
+            : base(string.Format(LocalizedStrings.InternalServerErrorMessage, message, exception))
+        { }
+
+        /// <summary>
+        /// This returns a properly formatted exception received from the server.
+        /// </summary>
+        public static InternalServerErrorException Create(string message, string exception)
+        {
+            var split = exception.Split(new[] { "\\r\\n", "\\r", "\\n" }, StringSplitOptions.RemoveEmptyEntries);
```
twsouthwick: You probably only need \\r and \\n, since \\r\\n will occur with that as well. Interesting thing I found: if you pass null in, it will split at all white space (looks like you only want to split at new lines, though); see the remarks at: https://msdn.microsoft.com/en-us/library/b873y76a.aspx

src/Microsoft.Fx.Portability/InternalServerErrorException.cs
```diff
@@ -0,0 +1,25 @@
...
+
+            return new InternalServerErrorException(message, string.Join(Environment.NewLine, split));
```
twsouthwick: Why are you removing newlines and then adding them back? is it to get a consistent newline character? Many times, this will be occurring on the service side, so Environment.NewLine will be from the server and may differ from what a client expects.

conniey:
Opting to add server side AppInsights logging for exceptions in lieu of exposing server side errors to customers.

PR #217, opened 2015-12-04
Create a separate updateable and updating cache

twsouthwick:
We had an implementation of an updating cache that would automatically update data. This separates the layering to have a cache that can be updated underneath a cache that automatically updates. This also exposes the .Start() method as public.


mjrousos:
UpdatingObjectCache.cs is showing up as a binary file in the diffs. Can you check its encoding to make sure it's UTF-8 or Unicode or something else standard that git recognizes?

twsouthwick:
Thanks - I just pushed a fix for that. It seems like we've got some encoding issues going on. Recent PRs keep having that happen...

src/Microsoft.Fx.Portability/Microsoft.Fx.Portability.NET45.csproj
```diff
@@ -17,14 +17,14 @@
     <DebugSymbols>true</DebugSymbols>
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
-    <DefineConstants>TRACE;DEBUG;NET45;FEATURE_ASSEMBLY_LOCATION;FEATURE_XML_SCHEMA</DefineConstants>
+    <DefineConstants>TRACE;DEBUG;NET45;FEATURE_ASSEMBLY_LOCATION;FEATURE_XML_SCHEMA;FEATURE_DEBUG_TRACE</DefineConstants>
```
mjrousos: I thought tracing APIs were available in .NET Core (in System.Diagnostics.TraceSource). Is that not the case?

twsouthwick: Hmmm... they may be now. I had disabled them when we first moved it over and they weren't available. They probably are now

src/Microsoft.Fx.Portability/Cache/AzureUrlBlobCache.cs
```diff
@@ -15,18 +15,16 @@ namespace Microsoft.Fx.Portability
     ///   -- When we are asked for the value of the cache we are going to return it.
     ///   -- Periodically we are going to check and see if the value of the blob is storage has changed (by using timestamps)
     /// </summary>
-    public class UrlBlobCache<TObject> : UpdatingObjectCache<TObject>
+    public class AzureUrlBlobCache<TObject> : UpdateableObjectCache<TObject>
```
mjrousos: Why did we make this an updateable cache instead of an updatingcache?

src/Microsoft.Fx.Portability/Cache/IObjectCache.cs
```diff
@@ -2,19 +2,15 @@
...
 {
     public interface IObjectCache : IDisposable
     {
-        Task UpdateAsync();
+        Task UpdateAsync(CancellationToken token);
 
         DateTimeOffset LastUpdated { get; }
     }
-
-    public interface IObjectCache<TObject> : IObjectCache
```
mjrousos: Was this interface not useful?

mjrousos: Never mind :) I see it in a separate file now.

src/Microsoft.Fx.Portability/Cache/AzureUrlBlobCache.cs
twsouthwick: Can't remember. I've switched it back.

mjrousos:
LGTM

PR #221, opened 2015-12-05
Switch to use CommonMark instead of MarkdownDeep

twsouthwick:
The main reason to do this is that MarkdownDeep.NET does not provide a nuget package that has a .NET45 assembly (only a NET35). By switching to use CommonMark, we can then reference the HTML reporter from project.json, allowing us to transition the service to ASP.NET 5.

This change also updates everything to use NuGet 3 - makes reasoning about dependencies easier.


tests/Microsoft.Fx.Portability.MetadataReader.Tests/ReflectionMetadataInfoTests.cs
```diff
@@ -46,7 +46,7 @@ public void UnresolvedAssemblyTest()
             Assert.Equal(s_expectedResult.Count(), actual.Count());
 
             // Use this instead of Assert.Equal so it will output the missing item
-            foreach (var items in s_expectedResult.Zip(actual, Tuple.Create))
+            foreach (var items in actual.Zip(s_expectedResult, Tuple.Create))
```
mjrousos: Why the change? I think that [Assert.Equal](https://github.com/xunit/xunit/blob/master/src/xunit.assert/Asserts/EqualityAsserts.cs) expects the 'expected' value first and I think [Zip](https://msdn.microsoft.com/library/dd267698%28v=vs.100%29.aspx) will take items from the calling collection as the 'first' ones, right?

tests/Microsoft.Fx.Portability.Tests/Microsoft.Fx.Portability.Tests.csproj
```diff
@@ -24,7 +21,7 @@
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
     <OutputPath>bin\Debug\</OutputPath>
-    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <DefineConstants>TRACE;DEBUG;FEATURE_XML_SCHEMA</DefineConstants>
```
mjrousos: Why is FEATURE_XML_SCHEMA defined? I don't see any changes here that require it.

tests/Microsoft.Fx.Portability.MetadataReader.Tests/ReflectionMetadataInfoTests.cs
twsouthwick: That's what I had thought, but it was failing here with some of the changes and I noticed the 'expected' and 'actual' were switched. Caused a bit of confusion, so I switched the parameters for next time it fails.

mjrousos: Weird. If it turns out that they really are switched, we may want to dig in more since there could be an xunit bug or something worth discovering.

tests/Microsoft.Fx.Portability.Tests/Microsoft.Fx.Portability.Tests.csproj
twsouthwick: I switched it to reference the desktop version and there is a test for TargetMapper that relies on that if/def. In desktop it validates the target map file against the schema and provides a slightly more helpful error message than the portable one where schema validation is not available.

mjrousos: Ah, I see.

mjrousos:
LGTM

twsouthwick:
Just a general comment as well about CommonMark.NET - it provides an awesome way to access the AST created during the conversion so we can hook in and change nodes: https://github.com/Knagis/CommonMark.NET/wiki

conniey:
LGTM. :)

PR #222, opened 2015-12-07
Override GetHashCode and Equals in BreakingChange class

twsouthwick:


tests/Microsoft.Fx.Portability.Tests/BreakingChangeTests.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    public class BreakingChangeTests
+    {
+        [Fact]
+        public void EqualityTests()
+        {
+            var breakingChangeSame1 = new BreakingChange { Id = "id1" };
+            var breakingChangeSame2 = new BreakingChange { Id = "id1" };
+            var breakingChangeDifferent1 = new BreakingChange { Id = "ID1" };
+            var breakingChangeDifferent2 = new BreakingChange { Id = "id2" };
```
mjrousos: Since the equality is case sensitive (which I think is probably correct), we should maybe add a test case to confirm that id1 and ID1 are not equal (just to make sure that behavior isn't inadvertently changed in the future).

mjrousos:
One nit-picky test suggestion; other than that LGTM

tests/Microsoft.Fx.Portability.Tests/BreakingChangeTests.cs
twsouthwick: I think I did - look at line 19 and line 28. Or are you referring to a specific test like:

```csharp
Assert.False(string.Equals(breakingChangeSame1.Id, breakingChangeDifferent1.Id, StringComparison.Ordinal));
```

mjrousos: Yep, somehow I missed line 19. Thanks!

mjrousos:
LGTM

PR #224, opened 2015-12-19
Add DiagnosticAnalyzerInfo and CompatibilityRangeAttribute classes

JJVertical:


src/Microsoft.Fx.Portability/ObjectModel/DiagnosticAnalyzerInfo.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public string AnalyzerName { get; set; }
+        public string Id { get; set; }
+        public ICollection<CompatibilityRangeAttribute> Attributes { get; set; }
+        public bool IsCompatibilityDiagnostic { get; set; }
+
+        public int GetIdNumber()
+        {
+            var matches = Regex.Match(Id, @"(\d+)");
+            return int.Parse(matches.Captures[0].Value);
```
twsouthwick: What about if someone creates an entry like this

```csharp
var info = new DiagnosticAnalyzerInfo { Id == "NoNumber" };
var id = info.GetIdNumber();
```

I'm pretty sure it'll fail. You should check whether a capture group was found.

src/Microsoft.Fx.Portability/ObjectModel/DiagnosticAnalyzerInfo.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class DiagnosticAnalyzerInfo
+    {
+        public string AnalyzerName { get; set; }
+        public string Id { get; set; }
+        public ICollection<CompatibilityRangeAttribute> Attributes { get; set; }
+        public bool IsCompatibilityDiagnostic { get; set; }
+
+        public int GetIdNumber()
+        {
+            var matches = Regex.Match(Id, @"(\d+)");
```
twsouthwick: You should pass in the RegexOptions.Compiled flag - this will likely be called many times.

src/Microsoft.Fx.Portability/Properties/version.txt
```diff
@@ -1 +1 @@
-1.2.2
+1.2.3
```
twsouthwick: Bump this to 1.3 - you're adding backwards compatible surface area

src/Microsoft.Fx.Portability/ObjectModel/DiagnosticAnalyzerInfo.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// Returns the integer number value from the Id string property or a -1 if an integer value cannot be returned. 
+        /// </summary>
+        public int GetIdNumber()
+        {
+            if (string.IsNullOrEmpty(Id))
+            {
+                return -1;
+            }
+
+            var matches = Regex.Match(Id, @"(\d+)", RegexOptions.Compiled);
```
mjrousos: If you're using RegexOptions.Compiled, it may make more sense to store the regex as a field in the class so that it can be re-used for each GetIdNumber call (instead of being re-generated). See [this](https://msdn.microsoft.com/en-us/library/yd1hzczs%28v=vs.110%29.aspx#Compiled) page about not letting compiled regexes go out of scope

twsouthwick: See the 3rd note - it still gets cached, although you can't manage its life so it may be removed from the cache. Making a static readonly field would be a better method to ensure it's only compiled once

src/Microsoft.Fx.Portability/ObjectModel/DiagnosticAnalyzerInfo.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+#if FEATURE_SERIALIZABLE
+    [Serializable]
+#endif
+    public class DiagnosticAnalyzerInfo
+    {
+        public string AnalyzerName { get; set; }
+        public string Id { get; set; }
+        public ICollection<CompatibilityRangeAttribute> Attributes { get; set; }
+        public bool IsCompatibilityDiagnostic { get; set; }
```
twsouthwick: Can you add a ICollection<string> Languages? That way we can also track which languages an analyzer supports.

src/Microsoft.Fx.Portability/ObjectModel/DiagnosticAnalyzerInfo.cs
mjrousos: Ah, good catch. I didn't realize they were cached (I suppose that makes sense). Given that, this matters less. It may still be nice (and a bit safer) to be explicit about the scope, though.

src/Microsoft.Fx.Portability/ObjectModel/CompatibilityRangeAttribute.cs
```diff
@@ -0,0 +1,20 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Fx.Portability.ObjectModel
+{
+#if FEATURE_SERIALIZABLE
+    [Serializable]
+#endif
+    public class CompatibilityRangeAttribute
```
conniey: Since this is no longer an attribute, you should rename it to `CompatibilityRange`.

src/Microsoft.Fx.Portability/ObjectModel/CompatibilityRangeAttribute.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Fx.Portability.ObjectModel
+{
+#if FEATURE_SERIALIZABLE
```
conniey: Question, why [Serializable]?

twsouthwick: We're using this type in the service to load information about the analyzers, which currently requires us to load in a separate app domain, necessitating the [Serializable] to send the information back

src/Microsoft.Fx.Portability/ObjectModel/CompatibilityRangeAttribute.cs
twsouthwick: :+1:

src/Microsoft.Fx.Portability/ObjectModel/CompatibilityRangeAttribute.cs
conniey: Ahh thanks for the explanation. :)

src/Microsoft.Fx.Portability/ObjectModel/DiagnosticAnalyzerInfo.cs
twsouthwick: The field Attributes should be changed to something like Ranges

src/Microsoft.Fx.Portability/ObjectModel/DiagnosticAnalyzerInfo.cs
```diff
@@ -0,0 +1,37 @@
...
+using System;
+
+namespace Microsoft.Fx.Portability.ObjectModel
+{
+#if FEATURE_SERIALIZABLE
+    [Serializable]
+#endif
+    public class DiagnosticAnalyzerInfo
+    {
+        private static readonly Regex _regEx = new Regex(@"(\d+)", RegexOptions.Compiled);
```
twsouthwick: nit: naming: s_regEx

src/Microsoft.Fx.Portability/Properties/version.txt
```diff
@@ -1 +1 @@
-1.2.2
+1.3
```
twsouthwick: Can you keep it as 1.3.0?

twsouthwick:
A couple of small changes, but otherwise LGTM

PR #225, opened 2015-12-30
Add another workaround link to 055

chlowell:
The MSDN breaking changes document doesn't mention the runtime configuration workaround.


mjrousos:
LGTM

PR #226, opened 2016-01-06
Fixing FxCop errors

conniey:
After running FxCop over the solution, several errors showed up.  This fixes errors:
- [CA1305: Specify IFormatProvider](https://msdn.microsoft.com/en-us/library/ms182190.aspx)
- [CA1307: Specify StringComparison](https://msdn.microsoft.com/en-us/library/bb386080.aspx)
- [CA2104: Do not declare read only mutable reference types](https://msdn.microsoft.com/en-us/library/ms182302.aspx)

Warning CA3053 is suppressed because the action required to fix it requires a field that does not exist in .NET Core.

### DIFF
--- a/src/Microsoft.Fx.Portability/BreakingChangeParser.cs
+++ b/src/Microsoft.Fx.Portability/BreakingChangeParser.cs
@@ -1,8 +1,10 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Microsoft.Fx.Portability.Resources;
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 
@@ -287,12 +289,12 @@ namespace Microsoft.Fx.Portability
                     // If a list of allowed categories was provided, make sure that the category found is on the list
                     if (!allowedCategories?.Contains(category, StringComparer.OrdinalIgnoreCase) ?? false)
                     {
-                        throw new InvalidOperationException($"Invalid category detected: {category}");
+                        throw new InvalidOperationException(string.Format(CultureInfo.CurrentCulture, LocalizedStrings.InvalidCategoryDetected, category));
                     }
                     currentBreak.Categories.Add(category);
                     break;
                 default:
-                    throw new InvalidOperationException("Unhandled breaking change parse state: " + state.ToString());
+                    throw new InvalidOperationException(string.Format(CultureInfo.CurrentCulture, LocalizedStrings.InvalidBreakingChangeParserState, state.ToString()));
             }
         }
 
--- a/src/Microsoft.Fx.Portability/DataExtensions.cs
+++ b/src/Microsoft.Fx.Portability/DataExtensions.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Fx.Portability
 {
     public static class DataExtensions
     {
-        public static readonly JsonSerializerSettings JsonSettings = new JsonSerializerSettings
+        public static JsonSerializerSettings JsonSettings { get; } = new JsonSerializerSettings
         {
             Formatting = Formatting.Indented,
             Converters = new JsonConverter[]
@@ -24,7 +24,7 @@ namespace Microsoft.Fx.Portability
             }
         };
 
-        public static readonly JsonSerializer Serializer = JsonSerializer.Create(JsonSettings);
+        public static JsonSerializer Serializer { get; } = JsonSerializer.Create(JsonSettings);
 
         public static byte[] Serialize<T>(this T data)
         {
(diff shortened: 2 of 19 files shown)


src/Microsoft.Fx.Portability/DataExtensions.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.Fx.Portability
 {
     public static class DataExtensions
     {
-        public static readonly JsonSerializerSettings JsonSettings = new JsonSerializerSettings
+        private static readonly JsonSerializerSettings JsonSettings = new JsonSerializerSettings
```
twsouthwick: Can you change this to a property? This is needed in the service.

twsouthwick: We need the settings in the service

src/Microsoft.Fx.Portability/ObjectModel/AnalyzeRequest.cs
```diff
@@ -37,7 +37,8 @@ public sealed class AnalyzeRequest : IComparable
         public int CompareTo(object obj)
         {
             var analyzeObject = obj as AnalyzeRequest;
-            return ApplicationName.CompareTo(analyzeObject.ApplicationName);
+
+            return string.Compare(ApplicationName, analyzeObject.ApplicationName, StringComparison.CurrentCulture);
```
conniey: I was hoping someone could comment on whether to use `StringComparison.CurrentCulture` or `StringComparison.Ordinal`.

src/Microsoft.Fx.Portability/DataExtensions.cs
```diff
@@ -24,7 +24,9 @@ public static class DataExtensions
             }
         };
 
-        public static readonly JsonSerializer Serializer = JsonSerializer.Create(JsonSettings);
+        private static readonly JsonSerializer JsonSerializer = JsonSerializer.Create(JsonSettings);
+
+        public static JsonSerializer Serializer { get { return JsonSerializer; } }
```
twsouthwick: Can you just do Serializer {get;} = JsonSerializer.Create(JsonSettings)?

src/Microsoft.Fx.Portability/ObjectModel/AnalyzeRequest.cs
twsouthwick: StringComparison.CurrentCulture (as anything that is culture specific) should be used when the strings are being displayed to the user. If it's internal to the program, ordinal should be used.

src/Microsoft.Fx.Portability/Reporting/ObjectModel/MissingTypeInfo.cs
```diff
@@ -54,7 +54,7 @@ public void IncrementUsage(AssemblyInfo sourceAssembly)
         public MissingTypeInfo(AssemblyInfo sourceAssembly, string docId, List<Version> targetStatus, string recommendedChanges)
             : base(docId)
         {
-            int pos = DocId.IndexOf("T:");
+            int pos = DocId.IndexOf("T:", StringComparison.OrdinalIgnoreCase);
```
twsouthwick: Is there a reason to ignore case? I think by the spec it needs to be capital

twsouthwick:
A couple of comments, otherwise LGTM

src/Microsoft.Fx.Portability/BreakingChangeParser.cs
```diff
@@ -287,12 +289,12 @@ private static void ParseNonStateChange(BreakingChange currentBreak, ParseState
                     // If a list of allowed categories was provided, make sure that the category found is on the list
                     if (!allowedCategories?.Contains(category, StringComparer.OrdinalIgnoreCase) ?? false)
                     {
-                        throw new InvalidOperationException($"Invalid category detected: {category}");
+                        throw new InvalidOperationException(string.Format(CultureInfo.CurrentCulture, LocalizedStrings.InvalidCategoryDetected, category));
```
conniey: [Interpolated strings](https://msdn.microsoft.com/en-us/library/dn961160.aspx) uses `string.Format(string, params object[])`, but to use this new feature with CultureInfo, the code needs to be changed to the example below. I wasn't sure which was the easiest approach to follow so I opted for the one above.

```csharp
var formattableString = (System.IFormattable)$"Invalid category detected: {category}"
throw new InvalidOperationException(formattableString.ToString(null, CultureInfo.CurrentCulture))
```

src/Microsoft.Fx.Portability/DataExtensions.cs
conniey: Fixed

src/Microsoft.Fx.Portability/DataExtensions.cs
conniey: Thanks for letting me know. I changed it to a static property

PR #227, opened 2016-01-06
Fix FxCop issues and `TypeInitializationException` when building

conniey:
- There were a lot more FxCop issues (after fixing the initial ones) when I tested it manually
   - [CA1305: Specify IFormatProvider](https://msdn.microsoft.com/en-us/library/ms182190.aspx)
   - [CA1307: Specify StringComparison](https://msdn.microsoft.com/en-us/library/bb386080.aspx)
   - Suppressed [CA2104](https://msdn.microsoft.com/en-us/library/ms182302.aspx) and CA5354
- Building from command-line resulted in a [Roslyn bug](https://github.com/dotnet/roslyn/issues/7276) which was fixed by upgrading to the newest nuget

### DIFF
--- a/src/ApiPort/CommandLine/ApiPortConfiguration.cs
+++ b/src/ApiPort/CommandLine/ApiPortConfiguration.cs
@@ -5,6 +5,7 @@ using Microsoft.Framework.Configuration;
 using Microsoft.Framework.OptionsModel;
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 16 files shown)


src/ApiPort/CommandLine/ApiPortConfiguration.cs
```diff
@@ -80,11 +81,12 @@ private static string ArrayTransform(string arg, ConfigMapCount array)
         {
             foreach (var tag in array.Tags)
             {
-                var newKey = $"{array.ExpectedTag}:{array.Count++}";
+                var newKey = string.Format(CultureInfo.InvariantCulture, "{0}:{1}", array.ExpectedTag, array.Count++);
```
twsouthwick: I think we should switch to the invariant culture, but take a look at this: http://www.thomaslevesque.com/2015/02/24/customizing-string-interpolation-in-c-6/. Creating a helper method Invariant or something might be a bit cleaner. I hate losing the string interpolation :(

conniey: I had tried this before but unfortunately `FormattableString` is only exposed .NET 4.6, and since our solution targets 4.5, it didn't work. :(

twsouthwick:
Ok, that's unfortunate FormattableString is not available. There is a workaround by Jon Skeet: https://gist.github.com/jskeet/9d297d0dc013d7a557ee. Your call if you want to implement that. Otherwise, LGTM

conniey:
Abandoning. Opting to disable FxCop for now.

PR #228, opened 2016-01-12
Add ability to get parent for an Api and fix OOM when serializing large objects

conniey:
- Add field to expose the parent docId for a given Api
- Encountered an OOM when serializing DotNetCatalog. Fix this by adding methods to write/read from stream rather than return a string.

### DIFF
--- a/src/Microsoft.Fx.Portability/DataExtensions.cs
+++ b/src/Microsoft.Fx.Portability/DataExtensions.cs
@@ -8,11 +8,16 @@ using System;
 using System.IO;
 using System.IO.Compression;
 using System.Runtime.Versioning;
+using System.Text;
+using System.Threading.Tasks;
 
 namespace Microsoft.Fx.Portability
 {
     public static class DataExtensions
     {
+        private const int DefaultBufferSize = 1024;
+        private static readonly Encoding DefaultEncoding = Encoding.UTF8;
+
         public static JsonSerializerSettings JsonSettings { get; } = new JsonSerializerSettings
         {
             Formatting = Formatting.Indented,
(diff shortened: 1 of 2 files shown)


src/Microsoft.Fx.Portability/ObjectModel/ApiInfoStorage.cs
```diff
@@ -13,6 +13,10 @@ public class ApiInfoStorage
         public string Type { get; set; }
         public string Name { get; set; }
         public string FullName { get; set; }
+        /// <summary>
```
twsouthwick: nit: new line above

src/Microsoft.Fx.Portability/ObjectModel/ApiInfoStorage.cs
```diff
@@ -13,6 +13,10 @@ public class ApiInfoStorage
         public string Type { get; set; }
         public string Name { get; set; }
```
twsouthwick: What's this type used for? Will changing the type break anything?

src/Microsoft.Fx.Portability/DataExtensions.cs
```diff
@@ -40,6 +45,20 @@ public static byte[] Serialize<T>(this T data)
...
         }
 
+        /// <summary>
+        /// Serializes an object to Json and writes the output to the given stream.
+        /// </summary>
+        /// <param name="data">object to serialize</param>
+        /// <param name="outputStream">Stream to write Json to</param>
+        /// <param name="leaveOpen">true to leave the stream open; false otherwise</param>
+        public static void Serialize<T>(this T data, Stream outputStream, bool leaveOpen)
+        {
```
twsouthwick: I don't see these methods being used anywhere...

conniey: These methods are used inside of our ApiCatalog exporter

src/Microsoft.Fx.Portability/ObjectModel/ApiInfoStorage.cs
conniey: ApiInfoStorage is a type that our service uses to read the raw data from the catalog. In `CloudApiCatalogLookup`, we end up converting this to `ApiDefinition`.. So the class not used directly.

src/Microsoft.Fx.Portability/ObjectModel/ApiInfoStorage.cs
conniey: fixed

twsouthwick:
LGTM

PR #232, opened 2016-01-15
Fixed nuget feed from dotnet-corefx to dotnet-core; update Metadata package

conniey:
- Move dotnet-corefx myget feed to dotnet-core
- Update Microsoft.CodeAnalysis because of TypeInitializationException in previous version
- Update System.Reflection.Metadata since it does not exist on the dotnet-core feed

### DIFF
--- a/src/Microsoft.Fx.Portability.MetadataReader/MemberMetadataInfoTypeProvider.cs
+++ b/src/Microsoft.Fx.Portability.MetadataReader/MemberMetadataInfoTypeProvider.cs
@@ -45,7 +45,7 @@ namespace Microsoft.Fx.Portability.Analyzer
                         Name = Reader.GetString(memberReference.Name),
                         ParentType = parentType,
                         Kind = MemberKind.Method,
-                        MethodSignature = SignatureDecoder.DecodeMethodSignature(memberReference.Signature, this).MakeEnclosedType()
+                        MethodSignature = memberReference.DecodeMethodSignature(this).MakeEnclosedType()
                     };
                 default:
                     return null;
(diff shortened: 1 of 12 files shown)


conniey:
@twsouthwick , @mjrousos 

One test is failing because it expects:
`M:TestClass.Bar(System.SByte optmod System.Runtime.CompilerServices.IsConst reqmod System.Runtime.CompilerServices.IsSignUnspecifiedByte*)`
but the only close to it in the actual results is:
`M:TestClass.Bar(System.SByte reqmod System.Runtime.CompilerServices.IsSignUnspecifiedByte optmod System.Runtime.CompilerServices.IsConst*)`

Original IL: [modopt.il](https://github.com/Microsoft/dotnet-apiport/blob/master/tests/Microsoft.Fx.Portability.MetadataReader.Tests/Tests/modopt.il#L23)

Is it possible the expected result is wrong?

twsouthwick:
I believe we got the expected result from CCI - may be another bug in the metadata reader

src/Microsoft.Fx.Portability.MetadataReader/MemberMetadataInfoTypeProvider.cs
```diff
@@ -395,19 +396,50 @@ public MemberMetadataInfo GetArrayType(MemberMetadataInfo elementType, ArrayShap
...
         {
             var builder = new StringBuilder();
             builder.Append(unmodifiedType.Name);
 
-            foreach (var modifier in modifiers)
+            builder.Append(isRequired ? " reqmod " : " optmod ");
+
+            MemberMetadataInfo info = null;
+
+            switch (modifierTypeHandle.Kind)
```
twsouthwick: Looks like the order in which this received is different than before

conniey:
Reflector shows: 
`public unsafe void Bar(sbyte modreq(IsSignUnspecifiedByte) modopt(IsConst)* x);`
CCI shows: 
`System.Void TestClass.Bar(System.SByte optmod System.Runtime.CompilerServices.IsConst reqmod System.Runtime.CompilerServices.IsSignUnspecifiedByte*)`

Maybe its a bug in CCI?

@nguerrera, Do you happen to have any insights?

twsouthwick:
@conniey Do we need to update the library at this point? Can we just get the feed update in so builds won't be broken?

conniey:
Yes. Because the old packages don't exist on the new feed. ;(

Cheers,
Connie

----- Reply message -----

twsouthwick:
Makes sense. I'm going to merge this in so we can get the builds working again. I have an issue open to track this now.

PR #238, opened 2016-01-20
Update a couple recommended changes

twsouthwick:
Now that @conniey got recommended changes coming from the repo, we can more easily update them.

### DIFF
--- a/docs/RecommendedChanges/System/Do not use.md
+++ b/docs/RecommendedChanges/System/Do not use.md
@@ -4,6 +4,3 @@ Do not use.  Couple behavior to feature availability, not OS version or Platform
 ### Affected APIs
 * `T:System.OperatingSystem`
 * `T:System.PlatformID`
-* `P:System.Environment.OSVersion`
-* `P:System.Environment.Version`
-* `M:System.Environment.get_OSVersion`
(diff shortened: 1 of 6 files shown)


twsouthwick:
@weshaggard Does the Framework Description recommended change apply to any other types?

weshaggard:
@Priya91 can you work with @twsouthwick get the recommeded changes updated for the RuntimeInformation API's you added.

Priya91:
@weshaggard Sure i'll follow up!

Priya91:
@twsouthwick 
- Use RuntimeInformation.ProcessArchitecture in place of System.Environment.Is64BitProcess
- Use RuntimeInformation.OSArchitecture in place of System.Environment.Is64BitOperatingSystem

```FrameworkDescription``` and ```OSDescription``` in ```RuntimeInformation``` are not replacements for ```Environment.Version``` or ```Environment.OSVersion```, since these API can be used for making runtime decisions whereas the RuntimeInformation description APIs are only for logging and display purposes.

weshaggard:
@Priya91 they don't have to be direct replacements but if we can point people to the new APIs that will help them satisfy some of their scenarios that is what we are after here in the recommended changes.

twsouthwick:
@Priya91 Keeping in mind that these are not 1:1 mappings as @weshaggard described, does this look right?

Priya91:
@twsouthwick Yes, LGTM!

PR #239, opened 2016-01-21
Fix ordering of signature modifiers to be in metadata order

### LINKED ISSUE #236: modopt doc id generation is broken
One test is failing because it expects:
`M:TestClass.Bar(System.SByte optmod System.Runtime.CompilerServices.IsConst reqmod System.Runtime.CompilerServices.IsSignUnspecifiedByte*)`
but the only close to it in the actual results is:
`M:TestClass.Bar(System.SByte reqmod System.Runtime.CompilerServices.IsSignUnspecifiedByte optmod System.Runtime.CompilerServices.IsConst*)`

Original IL: [modopt.il](https://github.com/Microsoft/dotnet-apiport/blob/master/tests/Microsoft.Fx.Portability.MetadataReader.Tests/Tests/modopt.il#L23)

Is it possible the expected result is wrong?
Reflector shows: 
 `public unsafe void Bar(sbyte modreq(IsSignUnspecifiedByte) modopt(IsConst)* x); `
 CCI shows: 
`System.Void TestClass.Bar(System.SByte optmod System.Runtime.CompilerServices.IsConst reqmod System.Runtime.CompilerServices.IsSignUnspecifiedByte*)`

Maybe its a bug in CCI?

Seen when upgrading from System.Reflection.Metadata 1.2.0-alpha-00017 to 1.2.0-rc2-23713 with PR #232

twsouthwick:
A recent fix in the beta of S.R.Metadata signature decoder returns the IL syntactic order instead of the metadata order we need for doc id generation. This will defer the name generation until we have all the modifiers.

Fixes #236

### DIFF
--- a/src/Microsoft.Fx.Portability.MetadataReader/MemberMetadataInfo.cs
+++ b/src/Microsoft.Fx.Portability.MetadataReader/MemberMetadataInfo.cs
@@ -17,6 +17,7 @@ namespace Microsoft.Fx.Portability.Analyzer
         {
             Kind = MemberKind.Type;
             Names = new List<string>();
+            Modifiers = new StackCollection<string>();
         }
 
         public MemberMetadataInfo(MemberMetadataInfo other)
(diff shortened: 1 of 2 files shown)


conniey:
LGTM :+1:

PR #243, opened 2016-01-29
Provide a ApiRecommendations class that searches the hierarchy from the catalog

twsouthwick:
This provides a class that can be used in offline mode and in the service that handles searching the hierarchy tree for recommended change/breaking changes/etc given a docid.


src/Microsoft.Fx.Portability/ObjectModel/AncestorApiRecommendations.cs
```diff
@@ -0,0 +1,86 @@
+﻿using System;
```
conniey: run CodeFormatter to add the license

conniey:
LGTM other than that one comment. :)

PR #270, opened 2016-02-29
Update build to use MicroBuild

conniey:
Updating our solution to use the new MicroBuild.  For now, removing ApiPort.Core from solution so that nuget.exe restore does not fail.

@twsouthwick , @mjrousos


twsouthwick:
Does NuGet 3.3 work with Core? That's the latest available right now.

conniey:
@twsouthwick NuGet 3.3 does work with Core (ie. If you execute `nuget.exe restore ApiPort.project.json`, it works successfully.) 

The problem is when we include it in our solution, it writes all of the resolved conflicting packages to the error stream.  Consequently, the build step fails because it saw that the error stream is not empty. Rather than ignoring all errors, I opted just to exclude these false ones until the issue is fixed.

twsouthwick:
Sounds good. LGTM

PR #272, opened 2016-03-02
Fixing version number for new build

conniey:


mjrousos:
LGTM

twsouthwick:
LGTM. Maybe leave a comment in there that to build on old-style XAML buids we need the TF_ version, while the new version doesn't have the prefix.

PR #275, opened 2016-03-07
Restoring packages in package directory in solution

conniey:
* Allows us to get the xUnit VS test adapter from a known location.

Reference: [How to run xunit test with vNext Build](http://www.donovanbrown.com/post/2015/06/15/how-to-run-xunit-test-with-vnext-build)

### DIFF
--- a/build/restore.ps1
+++ b/build/restore.ps1
@@ -1,7 +1,14 @@
 $root = $PSScriptRoot
+$packages = "$root\..\packages"
+
 $nuget = & "$root\Get-Nuget.ps1"
 
-& $nuget restore $root\..\PortabilityTools.sln
+New-Item $packages -ItemType Directory -ErrorAction Ignore
+
+# Restoring NuGet packages into this specific directory because the automated
+# build definition needs to be able to find the Visual Studio Test Adapter 
+# (xunit.runner.visualstudio) in a known location
+& $nuget restore $root\..\PortabilityTools.sln -PackagesDirectory $packages
 
 # Restore data for offline lib (cannot do this in the csproj due to race conditions)
 & $root\Get-CatalogFile.ps1 $root\..\.data\catalog.bin


build/restore.ps1
```diff
@@ -1,7 +1,14 @@
 $root = $PSScriptRoot
+$packages = "$root\..\packages"
+
 $nuget = & "$root\Get-Nuget.ps1"
 
-& $nuget restore $root\..\PortabilityTools.sln
+if (!(Test-Path $packages))
```
twsouthwick: Can you add a comment as to why we're doing this? Otherwise, we might remove it at some point, and then the tests start failing in automated builds

build/restore.ps1
```diff
@@ -1,7 +1,14 @@
...
+{
+    New-Item $packages -ItemType Directory
```
jasonwilliams200OK: We can use `mkdir $packages -ErrorAction Ignore`, to make this whole block one liner. :)

build/restore.ps1
conniey: Fixed. :)

build/restore.ps1
jasonwilliams200OK: FWIW, `mkdir` is the alias of `New-Item -ItemType Directory` since XP SP3. :)

twsouthwick: The recommendations I've seen for powershell scripts is to use built-in names, not aliases, in case they have been changed to something else.

jasonwilliams200OK: Ah sorry, mkdir is a command itself since XP. The alias of New-Item is `ni`.
mkdir is a DOS command, which means it was there for 30+ years :smile:

PowerShell let us add `-ErrorAction Ignore` to any command, which has the same effect as `-p` switch of mkdir command in Unix shells.

twsouthwick: I didn't realize -ErrorAction Ignore could be added to any command - very cool to know!

twsouthwick:
LGTM

PR #277, opened 2016-03-22
Recommended Change for WebProxy

SidharthNabar:
Added a recommended change for the WebProxy type


twsouthwick:
Thanks @SidharthNabar

PR #283, opened 2016-04-07
Add code sample for Timer.Dispose(WaitHandle) replacement

dsplaisted:
This adds a code sample for what you can do to replace calls to Timer.Dispose(WaitHandle).

Is it a good idea to put code samples like this in the recommended changes file itself?  Or should we keep it somewhere else and link to it?

### DIFF
--- a/docs/RecommendedChanges/System.Threading/Use Monitor.md
+++ b/docs/RecommendedChanges/System.Threading/Use Monitor.md
@@ -4,7 +4,7 @@ If you need to synchronize destruction of the timer with execution of the callba
 ### Affected APIs
 * `M:System.Threading.Timer.Dispose(System.Threading.WaitHandle)`
 
-### Replacement code
+### Replacement Code
 The following class shows how you can wait for the callback of a timer to finish running. 
 
 ```C#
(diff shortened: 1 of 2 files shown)


dsplaisted:
cc @ericeil

ericeil:
The code sample looks good; I'm not sure how the overall format of the document fits with the rest of apiport, so I can't comment on that.

docs/RecommendedChanges/System.Threading/Use Monitor.md
```diff
@@ -1,5 +1,78 @@
 ### Recommended Action
-If you need to synchronize destruction of the timer with execution of the callback, do this yourself via Monitor or other synchronization mechanism.  If not, simply use Dispose().
+If you need to synchronize destruction of the timer with execution of the callback, do this yourself via Monitor or other synchronization mechanism.  If not, simply use Dispose().  Code sample: http://aka.ms/porttocore_timer_dispose
 
 ### Affected APIs
 * `M:System.Threading.Timer.Dispose(System.Threading.WaitHandle)`
+
+### Replacement code
```
conniey: Title case the header: Replacement Code

conniey:
Would you be able to add this section into the [! Template.md](docs/RecommendedChanges/!%20Template.md) as well and make a note that it's an optional section?

The format of the document looks OK. We only parse out the "Recommended Actions" and 
"Affected APIs" in the .NET Portability analyzer. So they won't see the "Replacement Code" section in their output.

twsouthwick:
@dsplaisted Once you update the `! Template.md` file, this can be merged in. Since it's a new section, it would be great to get it documented.

dsplaisted:
@twsouthwick OK, I've updated the `! Template.md` file with the new section.  Let me know what you think!

twsouthwick:
Perfect! LGTM

PR #288, opened 2016-04-15
Add field for descriptions in target

### LINKED ISSUE #287: Provide descriptions for targets
We should add a way to describe what each target is for that can then be surfaced to users.

twsouthwick:
This also updates the listTargets command to output the description received from the service

@conniey 

Fixes #287.


conniey:
LGTM :+1:

PR #289, opened 2016-04-19
Switch ApiPort to autofac instead of unity

### LINKED ISSUE #211: Use Autofac for dependency injection in ApiPort instead of Unity
This will allow easy run-time configuration in .NET Core scenarios (which don't support System.Configuration).

twsouthwick:
Autofac has better .NET Core support and so we can enable all features on Core. I switched the projects to use the dotnet CLI in another branch and ran up against this again.

@mjrousos  @conniey 

Fixes #211


src/Microsoft.Fx.Portability.Offline/OfflineDataModule.cs
```diff
@@ -0,0 +1,61 @@
...
+            builder.RegisterType<OfflineApiRecommendations>()
+                .As<IApiRecommendations>()
+                .SingleInstance();
+
+            builder.RegisterType<JsonReportWriter>()
+                .As<IReportWriter>()
+                .SingleInstance();
+
+#if FEATURE_HTML_WRITER
+            // Currently, the HTML writer has dependencies that do not work on .NET Core
```
mjrousos: Which parts of the HtmlWriter don't work in Core?

mjrousos:
LGTM

src/Microsoft.Fx.Portability.Offline/OfflineDataModule.cs
twsouthwick: RazorEngine - at least the version we're using.

PR #293, opened 2016-04-26
Add recommendation for X509Certificate2.PrivateKey

mjrousos:
@bartonjs

### DIFF
--- a/docs/RecommendedChanges/System.Security.Cryptography.X509Certificates/Use GetRSAPrivateKey or GetECDsaPrivateKey.md
+++ b/docs/RecommendedChanges/System.Security.Cryptography.X509Certificates/Use GetRSAPrivateKey or GetECDsaPrivateKey.md
@@ -1,0 +1,5 @@
+### Recommended Action
+Use GetRSAPrivateKey() or GetECDsaPrivateKey() extension methods instead.
+
+### Affected APIs
+* `P:System.Security.Cryptography.X509Certificates.X509Certificate2.PrivateKey`


docs/RecommendedChanges/System.Security.Cryptography.X509Certificates/Use GetRSAPrivateKey or GetECDsaPrivateKey.md
```diff
@@ -0,0 +1,5 @@
+### Recommended Action
+Use GetRSAPrivateKey() or GetECDsaPrivateKey() extension methods instead.
```
twsouthwick: Is this in the same assembly or does one need to be added?

mjrousos: In .NET Core, the extension is in the same contract so it should be easy to find. Curiously, on desktop .NET the extension is in System.Core whereas the rest of X509Certificate2 is in System.

twsouthwick: Ok - that's probably fine. We had a problem with some recommended changes that recommended an extension method that required adding a separate package and a user couldn't find it.

twsouthwick:
LGTM

bartonjs:
LGTM

PR #296, opened 2016-04-26
Allow package restore location to be set in a way that gets propagated

### LINKED ISSUE #295: Builds fail when NuGet cache is cleared
Looks like a change (#275) to enable builds to run xUnit in VSTS fails when the NuGet cache is cleared.

twsouthwick:
@conniey 

Fixes #295

### DIFF
--- a/build/restore.ps1
+++ b/build/restore.ps1
@@ -1,14 +1,17 @@
 $root = $PSScriptRoot
-$packages = "$root\..\packages"
-
 $nuget = & "$root\Get-Nuget.ps1"
 
-New-Item $packages -ItemType Directory -ErrorAction Ignore
-
 # Restoring NuGet packages into this specific directory because the automated
-# build definition needs to be able to find the Visual Studio Test Adapter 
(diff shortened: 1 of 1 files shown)


conniey:
Lgtm

build/restore.ps1
```diff
@@ -1,14 +1,17 @@
...
 $nuget = & "$root\Get-Nuget.ps1"
 
-New-Item $packages -ItemType Directory -ErrorAction Ignore
-
 # Restoring NuGet packages into this specific directory because the automated
-# build definition needs to be able to find the Visual Studio Test Adapter 
+# build definition in VSTS needs to be able to find the Visual Studio Test Adapter 
 # (xunit.runner.visualstudio) in a known location
-& $nuget restore $root\..\PortabilityTools.sln -PackagesDirectory $packages
+if($env:NuGetPackagesDirectory)
```
conniey: Interesting! When is that env variable set?

twsouthwick: You'll need to update the build def to set it, but it seems to be used deep in the msbuild scripts to determine where to resolve packages.

PR #298, opened 2016-04-28
Use a concurrent set to track user assemblies

### LINKED ISSUE #297: Adding user assemblies can create a race condition
`ReflectionMetadataDependencyInfo` adds assemblies to a `HashSet<>` in parallel without locking. This creates a race condition where for large sets of assemblies adding another item can cause the add method to explode.

twsouthwick:
Fixes #297

### DIFF
--- a/src/Microsoft.Fx.Portability.MetadataReader/ReflectionMetadataDependencyInfo.cs
+++ b/src/Microsoft.Fx.Portability.MetadataReader/ReflectionMetadataDependencyInfo.cs
@@ -4,6 +4,7 @@
 using Microsoft.Fx.Portability.Analyzer.Resources;
 using Microsoft.Fx.Portability.ObjectModel;
 using System;
+using System.Collections;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Fx.Portability.MetadataReader/ReflectionMetadataDependencyInfo.cs
```diff
@@ -196,5 +197,33 @@ public InvalidPEAssemblyException(Exception inner)
                 : base("Not a valid assembly", inner)
             { }
         }
+
```
conniey: Have you considered using (ImmutableHashSet<T>)[https://msdn.microsoft.com/en-us/library/dn467171(v=vs.110).aspx]?

twsouthwick: Yup - I prototyped that as well, but then you have to deal with Interlocked.Exchange or something to ensure the updated set is swapped with the original since immutable collections don't update. I went with this as we're already using ConcurrentDictionaries and it was a fairly simple swap.

conniey: You do have to use `ImmutableInterlocked.Update<T>(ref T location, Func<T, T> transformer)` to update the set. The benefit is that this would be explicit in telling users that the collection does not change (except upon initialization).

I would also consider changing `_unresolvedAssemblies` and `_cachedDependencies` to `ImmutableDictionary`.

It's a nice to have. The PR looks good. :)

twsouthwick: Cool - I'll take a look at that. I would prefer the immutable collections for the guarantees they provide.

twsouthwick: I replaced everything with ImmutableCollections and then did some checks around performance. Looks like using ImmutableCollections here will decrease performance by about 40% - we're doing all the building initially, which is known to drag performance with them. I think we should stick with the concurrent collections for now.

conniey:
LGTM. :+1:

PR #315, opened 2016-08-27
Adding support for .NETStandard 1.3

conniey:
This PR:
* Adds support for .NET Standard
* Updates Newtonsoft.Json to fix bug in Azure WebApp Json deserialisation errors

@twsouthwick Can you check the updates I made to the `version.txt`s? I updated the versions based on the changes:
* Increased major version if the target framework changed from `dotnet` to `netstandard1.3`
* Increased minor version if only the dependencies were updated


twsouthwick:
This looks fine, except looks like some invalid docids were found. If you build it on your machine, do the resulting nuget packages work correctly?

conniey:
@terrajobst Is the API Catalog capable of identifying methods in a class when they've been overridden from its base class?

**Example**
The catalog says `M:System.Windows.Threading.DispatcherSynchronizationContext.CreateCopy` does not exist, but I can see it in the [MSDN documentation](https://msdn.microsoft.com/en-us/library/system.windows.threading.dispatchersynchronizationcontext(v=vs.110).aspx#Anchor_3).

Our checks are failing because of this. :/ Can you suggest a workaround?

terrajobst:
We've recently made a change to the API Catalog to no longer index overrides and explicit interface implementations as they result in issues when we use the catalog to diff two platforms.

What check are you running?

conniey:
@terrajobst I thought it was affecting the analyses but it turns out, our check is looking for the docIds as-is, rather than considering that it may be inherited from a base class. A fix on our end. :) Thanks for the information!

terrajobst:
That's what I thought. Thanks for following up :-)

PR #319, opened 2016-09-27
Merge from devdiv

twsouthwick:
The biggest change this brings is the VS extension code.


src/ApiPort.VS2013/Analyze/ApiPortVsAnalyzer.cs
```diff
@@ -0,0 +1,198 @@
+﻿using ApiPortVS.Resources;
```
conniey: Need to run the CodeFormatter tool to add the MIT header to the *.cs files.

src/ApiPort.VS2013/ApiPort.VS2013.csproj
```diff
@@ -0,0 +1,325 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
conniey: Rename the project from ApiPort.VS2013 back to ApiPortVS since does not open on VS2013 anymore.

conniey: Or would it be better done in another PR?

twsouthwick: Done

src/ApiPort.VS2013/Analyze/ApiPortVsAnalyzer.cs
twsouthwick: Done

src/ApiPort.Vsix/ApiPort.Vsix.sln
```diff
@@ -0,0 +1,50 @@
+﻿
```
conniey: We don't need this file anymore since it's integrated into PortabilityTools.sln

twsouthwick: Sounds good. I just saw that file and was wondering what its use was... I've removed it now.

PR #332, opened 2016-10-01
Update ApiPort.Core to .NET Core App

conniey:
* Update ApiPort.Core to .NETCoreApp
* Update README


src/ApiPort/ApiPort.csproj
```diff
@@ -93,9 +93,6 @@
     </EmbeddedResource>
   </ItemGroup>
   <ItemGroup>
-    <None Include="app.config">
```
twsouthwick: Why remove this?

PR #334, opened 2016-10-03
Migrate test projects to xUnit

conniey:
* Migrates the following projects to xUnit from VSTest
   * ApiPortVS.Test
   * Microsoft.Fx.Portability.Cci
   * Microsoft.Fx.Portability.Offline
* `TestAssembly` filenames are now unique because xUnit tests run in parallel and there was often exceptions being thrown in the Cci tests because another test was using the same assembly
* `TargetPlatform.CompareTo` takes into consideration the Versions as well because xUnit's `Assert.Equal` was failing due to these lines:
   * [AssertEqualityComparer: L60 - 62](https://github.com/xunit/assert.xunit/blob/master/Sdk/AssertEqualityComparer.cs#L60-L62) checks that if a class inherits from `IComparable<T>` and returns 0, then the objects are equal. It would call the `object.Equals` override only if it had exhausted all other options ([Line 85](https://github.com/xunit/assert.xunit/blob/master/Sdk/AssertEqualityComparer.cs#L85)).


src/ApiPort.VisualStudio/TargetPlatform.cs
```diff
@@ -89,9 +88,55 @@ public override int GetHashCode(TargetPlatform obj)
...
+            var comparedNames = string.CompareOrdinal(Name, other.Name);
+
+            if (comparedNames != 0)
+            {
+                return comparedNames;
+            }
+
+            // Adapted from:
+            // https://github.com/dotnet/corefx/blob/master/src/System.Linq/src/System/Linq/SequenceEqual.cs#L43-L55
+            using (var e1 = Versions.GetEnumerator())
```
twsouthwick: This seems like overkill - why not just use SequenceEquals?

conniey: SequenceEquals would only tell you if your response to CompareTo should be 0. But I'd need to actually compare the items inside the list to see whether or not it was -1 or 1.

twsouthwick: Ah - can you add a comment to that effect?

conniey: kk Done.

PR #337, opened 2016-10-05
Fixing build and tests

conniey:
@twsouthwick


twsouthwick:
Thanks! Not sure how it was building on my machine... I probably didn't stage things correctly.

PR #338, opened 2016-10-05
Add recommended change for CharSet.Auto

### LINKED ISSUE #299: CharSet.Auto is missing from NetCore
https://msdn.microsoft.com/en-us/library/system.runtime.interopservices.charset.aspx

In netcore, `Auto` is not present in `CharSet` enum, we only have `Ansi`, `Unicode` and `None`.

twsouthwick:
Fixes #299


twsouthwick:
@yizhang82 Does this look right?

PR #344, opened 2016-10-07
Remove .NET4.5 projects

conniey:
Since ApiPort supports .NET Core App now, we can update the ApiPort desktop project to .NET 4.6.  This PR:

* Removes .NET 4.5 projects because ApiPort can reference the .NET Standard 1.3 projects
* Moves ApiPort libraries out of nested solution folders
* Fixes crashes in ApiPort.Offline due to mismatched System.Reflection.Metadata (from the copy order in postbuild.ps1)


conniey:
> let's make sure the .NET Core build runs on Win7 x32 and x64 

I ran it on Windows 7 Enterprise N (x64) with SP1 but I couldn't find an image in Azure to try out x32.

twsouthwick:
Ok - I think xlab would have one. But that's probably fine for now.

conniey:
I tested it on Windows 7 x86 and it works. :)

PR #348, opened 2016-10-10
Adding ability to publish VSIX

conniey:
* Allows our CI build to publish updates to [vsixgallery](http://vsixgallery.com/)
* Fixes nuget warnings while packaging nuget package because libraries are not stable releases


build/postbuild.ps1
```diff
@@ -83,18 +101,29 @@ function Copy-OfflineMode()
...
\ No newline at end of file
+Copy-Item "$root\..\docs\LicenseTerms" $drop\ApiPort\$netFramework -Recurse -Force
+Copy-Item "$root\..\docs\LicenseTerms" $drop\ApiPort.Offline\ -Recurse -Force
+
+if ($PublishVsix) {
+    # Setting these environment variables because they are used in the script
+    # for the upload endpoint:
+    # https://github.com/madskristensen/ExtensionScripts/blob/master/AppVeyor/vsix.ps1#L52-L55
+    # https://github.com/madskristensen/ExtensionScripts/blob/master/AppVeyor/vsix.ps1#L68
+    $env:APPVEYOR_REPO_PROVIDER = "github"
```
twsouthwick: It may be good to just grab that script so it doesn't change underneath us

twsouthwick:
Are you handling version numbering for CD builds? Right now, versions are taken from the version.txt files, but to do CD, we would want at least the revision number changing on each deployment

build/postbuild.ps1
conniey: In [build/Get-BuildTools.ps1:L28-30](https://github.com/conniey/dotnet-apiport/blob/4fe2ceb6a3e87365b1856f417413f58113711ca1/build/Get-BuildTools.ps1#L27-L30), I get it from the github repository and dump it in .tools. Is that what you were asking?

conniey:
> Are you handling version numbering for CD builds? Right now, versions are taken from the version.txt files, but to do CD, we would want at least the revision number changing on each deployment

We don't use version.txt for ApiPort.Vsix builds. But, in [restore.ps1](https://github.com/conniey/dotnet-apiport/blob/4fe2ceb6a3e87365b1856f417413f58113711ca1/build/restore.ps1#L8), we increment the "revision" number by default, so takes into account the version that exists in the vsix manifest.

twsouthwick:
We should consolidate the versioning strategy, but that'll work for now

PR #349, opened 2016-10-10
Minimize requests during analysis for multiple output formats

twsouthwick:
A recent change to the service will allow output as multipart if multiple formats were requested. This change modifies the ApiPortClient and ApiPortService class to use that to minimize output formats. As it is, if someone requests multiple formats, the entire payload is submitted multiple times; with this change, it will only be submitted once.


src/Microsoft.Fx.Portability/CompressedHttpClient.cs
```diff
@@ -93,47 +106,118 @@ public CompressedHttpClient(ProductInformation info)
                 FileExtension = ".json"
             }; ;
 
-            var response = await CallInternalAsync(request, json);
-            var result = response.Response.Deserialize<TResponse>();
+            var response = await CallInternalAsync(request, new[] { json });
+            var result = response.Response.FirstOrDefault().Data.Deserialize<TResponse>();
```
twsouthwick: I've switched this to just use `.Single()` since there should be 1 and only 1 in the enumerable. This is because the length of the array passed to `CallInternalAsync` should be the same as the output. The `OrDefault` ones should be used if there's a chance it will be an empty enumerable; that is not the case with this.

twsouthwick:
As far as upping the semantic version, technically, we're still in preview, so we can make breaking changes. We should look at releasing a stable version of the libraries... I'll update the minor version so we don't get conflict.

conniey:
Thanks for explaining. :) Versioning is always a confusing thing. LGTM!

PR #354, opened 2016-10-12
Migrating to use project.json instead of packages.config

conniey:
Fixes errors in VSIX because ApiPort.Vsix was using packages.config and Microsoft.Fx.Portability uses project.json.  As a result, the correct dependencies were not being copied over.  So, at runtime, we would get a FileNotFoundException because a dependency in MS.Fx.Portability could not be resolved.

/cc @twsouthwick


conniey:
> just make sure everything shows up in the VSIX with this change. I've had problems with that in the past

That was the motivation for moving to use project.json. I was Dogfooding the VSIX and it was failing due to the missing dependencies.

Issue #355, opened 2016-10-13
Visual Studio Extension Error


modulexcite:
Hi,

Thank you for the extension.
The new version of the extension 1.3.0.114 downloaded from VSIX gallery is throwing an error "could not load file or assembly Autofac" and it seems like the packaged vsix is missing a few dependent libraries such as Autofac, etc.
I'm using Visual Studio 2015 Update 3, the older version v1.2 of this extension from the MSDN gallery seems to work fine and has the dependent assemblies packaged with it.
Please let me know if you need more information.

conniey:
Thanks for letting us know. Surprised that we have folks Dogfooding that CI build already! I am working on fixing this issue. It was due to my PR #354 . :/

modulexcite:
Guilty:)
Thought I'd get on the new features bandwagon for this extension before it bit me in the ass:)

___________________________
Thanks.




On Thu, Oct 13, 2016 at 11:52 AM -0400, "Connie Yau" <notifications@github.com> wrote:












Thanks for letting us know. Surprised that we have folks Dogfooding that CI build already! I am working on fixing this issue. It was due to my PR #354 . :/

modulexcite:
Thanks for fixing it by the way:)

twsouthwick:
@connie just saw this issue, but I think a PR I just submitted will fix this issue. By switching to project.json, the NuGet dependencies were not being included by default into the VSIX. I found a work-around for now.

twsouthwick:
@modulexcite can you update to the latest version and give it a try?

twsouthwick:
You'll need 1.3.0.115 or later

modulexcite:
Has it been published to the VSIX gallery?

modulexcite:
Never mind, see it on the gallery.

modulexcite:
Now i get a different error upon running the analyzer, see below. I failed to mention earlier that I'm on a Windows 7 SP1 OS with VS 2015 Professional Update 3.
***
Analyzing compatibility                            [Failed]

There was an unknown error code from the service: 'InternalServerError'.  Please make sure you have the most up-to-date version.

conniey:
@modulexcite Weird. It may be because a few of our last builds had the incorrect digital signature. We fixed it in version 1.3.0.117 (the latest one on the vsixgallery.com). 

In the meantime, I'll spin up a Windows 7 VM and try to repro your issue.

modulexcite:
[...]

PR #356, opened 2016-10-13
Update to NETStandard.Library and clean up VSIX redirection

twsouthwick:
This change fixes up some inconsistencies with references:

- project.json does not embed interop types in VSIX; added helper NuGet package for this
- The projects had been using Microsoft.NETCore instead of NETStandard.Library
- NuGet references were not being included in VSIX. A custom targets file was added to fix up the references to be included. This is a manual step, but there may be a way to automate it
- There were extra assembly redirects that are not needed


twsouthwick:
This will fix #355 as well

PR #364, opened 2016-10-17
Add menu option to analyze transitive project references

### LINKED ISSUE #363: Add option to analyze all output in VSIX
Right now there are three options:

1. Analyze solution
2. Analyze supplied assemblies
3. Analyze individual project

We should add an option for individual project to analyze all dependent projects as well

twsouthwick:
Fixes #363

### DIFF
--- a/src/ApiPort.VisualStudio/AnalyzeMenu.cs
+++ b/src/ApiPort.VisualStudio/AnalyzeMenu.cs
@@ -31,17 +31,19 @@ namespace ApiPortVS
             _output = output;
         }
 
-        public async void ProjectContextMenuItemCallback(object sender, EventArgs e)
+        public async Task AnalyzeSelectedProjectsAsync(bool includeDependencies)
         {
-            await AnalyzeProjects(GetSelectedProjects());
+            var projects = GetSelectedProjects();
+
+            await AnalyzeProjectsAsync(includeDependencies ? GetTransitiveReferences(projects, new HashSet<Project>()) : projects);
         }
 
         public async void SolutionContextMenuItemCallback(object sender, EventArgs e)
         {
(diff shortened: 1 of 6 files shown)


src/ApiPort.VisualStudio/AnalyzeMenu.cs
```diff
@@ -31,9 +31,11 @@ public AnalyzeMenu(ILifetimeScope scope, DTE dte, TextWriter output)
             _output = output;
         }
 
-        public async void ProjectContextMenuItemCallback(object sender, EventArgs e)
+        public async Task AnalyzeSelectedProjects(bool includeDependents)
```
conniey: nit: AnalyzeSelectedProjectsAsync

src/ApiPort.VisualStudio/AnalyzeMenu.cs
```diff
@@ -171,5 +184,23 @@ private ICollection<Project> GetSelectedProjects()
                 .Distinct()
                 .ToList();
         }
+
+        private ICollection<Project> GetTransitiveReferences(IEnumerable<Project> projects, HashSet<Project> expandedProjects = null)
```
conniey: If the default case is `expandedProjects = null`, when does that get initialized? Is there a NullReferenceException when this method is called?

twsouthwick: Look a couple of lines down. Since this is recursive, the initial call doesn't need to supply this, but it is checked the second line down. I had originally had the caller supply a the hashset - I'll switch it back to that so it's clearer.

twsouthwick: Good catch - I must have changed this without trying it out. I was checking the wrong variable! I'm definitely switching it back to what I had

PR #368, opened 2016-10-18
Remove custom assembly redirects logic

twsouthwick:
VS provides the ProvideBindingRedirection that can handle our binding redirect needs.


twsouthwick:
@conniey Can you take a look at this?

PR #369, opened 2016-10-19
Make public CloudApiCatalogLookup members virtual

chlowell:
This enables consumers, e.g. the portability service, to implement target-specific lookup logic.

### DIFF
--- a/src/Microsoft.Fx.Portability/ObjectModel/CloudApiCatalogLookup.cs
+++ b/src/Microsoft.Fx.Portability/ObjectModel/CloudApiCatalogLookup.cs
@@ -65,7 +65,7 @@ namespace Microsoft.Fx.Portability.ObjectModel
             _docIdToApi = catalog.Apis.ToDictionary(key => key.DocId, key => new ApiDefinition { DocId = key.DocId, Name = key.Name, ReturnType = key.Type, FullName = key.FullName, Parent = key.Parent });
         }
 
-        public IEnumerable<string> DocIds
+        public virtual IEnumerable<string> DocIds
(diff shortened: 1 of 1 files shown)


twsouthwick:
I don't think we want to hard code what we union. Is there a way to abstract this so we can configure it run time?

chlowell:
If I understand the status quo, `TargetMapper` can't be used for this because it only enables simple aliasing of target names. After mapping "ASP.NET Core" to ".NET Core App", analyzing against the former target would be equivalent to analyzing against the latter (correct me if I'm wrong here). That won't union the targets' API surfaces.

I can look at adding a unioning feature to `TargetMapper`, enabling `TargetMap.xml` to specify unions at runtime.

twsouthwick:
That's correct. TargetMapper has similar functionality so it may be a good place to start. Probably should create a new interface for it, though

conniey:
@chlowell Is this still necessary since we've moved it to the servivce?

chlowell:
What's in the service is a temporary expedient to get the change into production quickly. It's the code here plus a base class copied from the Portability assembly--definitely intended to be temporary. But overall this is a hacky way to solve the problem and I'm working on another approach. I'll close this PR since I'm not moving forward with these changes (though I might return to target mapping).

twsouthwick:
Can you at least get the changes in that make the methods virtual so we don't need to keep a copy in the service?

chlowell:
Good idea. I've updated this to make only those changes.

twsouthwick:
Can you update all the public methods to be virtual as well?

PR #370, opened 2016-10-22
Minor changes to the Introduction topic

mairaw:
I just noticed that the Framework wasn't called out in the heading and improved other little things I saw.


twsouthwick:
LGTM, thanks!

PR #376, opened 2016-11-17
Add support to hide columns in the portability HTML report

jeremymeng:
/cc @twsouthwick @conniey

### DIFF
--- a/src/Microsoft.Fx.Portability.Reports.Html/Resources/_PortabilityReport.cshtml
+++ b/src/Microsoft.Fx.Portability.Reports.Html/Resources/_PortabilityReport.cshtml
@@ -106,63 +106,77 @@
     </tbody>
 </table>
 <div id="details">
-    @foreach (var assembly in Model.OrderedAssembliesByIdentity)
-    {
-        // avoid writing blank details tables (happens if a non-.NET assembly is analyzed)
-        bool isDotNETAssembly = false;
-        foreach (var type in Model.MissingTypes)
-        {
-            if (type.UsedIn.Contains(assembly.SourceAssembly))
+        @{ 
+            var detailColumnHeaders = new List<string>();
+            detailColumnHeaders.Add(LocalizedStrings.TargetTypeHeader);
+            detailColumnHeaders.AddRange(Model.TargetHeaders);
+            detailColumnHeaders.Add(LocalizedStrings.RecommendedChanges);
+        <h3>Hide columns:</h3>
+        <form>
+            @* It doesn't make sense to hide the first column (Target type), which lists
+                names of the types and members that have possible problems in porting. *@
+            @for (int i = 1; i < detailColumnHeaders.Count; i++)
             {
-                isDotNETAssembly = true;
-                break;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Fx.Portability.Reports.Html/Resources/_PortabilityReport.cshtml
```diff
@@ -144,7 +144,19 @@
         detailColumnHeaders.Add(LocalizedStrings.TargetTypeHeader);
         detailColumnHeaders.AddRange(Model.TargetHeaders);
         detailColumnHeaders.Add(LocalizedStrings.RecommendedChanges);
-        <table>
+
+        <form>
+            Hide columns: 
+            @* It is not allowed to hide the first (Api name) and the last (Recommended Changes) columns. *@
```
twsouthwick: Does this mean that we don't want them to or it's not possible?

jeremymeng:
Here's what it looks like after the change:
image

src/Microsoft.Fx.Portability.Reports.Html/Resources/_Scripts.cshtml
```diff
@@ -40,4 +40,32 @@
         var currentClass = beginCollapsed[i].getAttribute('class');
         beginCollapsed[i].setAttribute('class', currentClass.concat(' ToggledOff').trim());
     }
+
+    // Toggle the visibility of a column in the portability report detail
+    // table (id='table_portability_details')
+    function show_hide_column(col_no, do_show) {
+
```
twsouthwick: nit: newline

twsouthwick:
Looks like the change introduces the selection per-assembly. I think it would make more sense to have it for the whole report

src/Microsoft.Fx.Portability.Reports.Html/Resources/_PortabilityReport.cshtml
jeremymeng: We don't allow as it wouldn't make sense to hide the first column.  Maybe allow hiding the last?

twsouthwick:
Can you show an updated screenshot with the changes?

jeremymeng:
updated screenshot:

image

src/Microsoft.Fx.Portability.Reports.Html/Resources/_PortabilityReport.cshtml
```diff
@@ -106,63 +106,78 @@
...
 </table>
 <div id="details">
-    @foreach (var assembly in Model.OrderedAssembliesByIdentity)
-    {
-        // avoid writing blank details tables (happens if a non-.NET assembly is analyzed)
-        bool isDotNETAssembly = false;
-        foreach (var type in Model.MissingTypes)
+        @{ 
+            var detailColumnHeaders = new List<string>
+            ();
```
conniey: Move it to the line above into `new List<string>();`

jeremymeng:
Thanks!  I add a small bug fix and squashed the changes.

jeremymeng:
Thanks!

Issue #261, opened 2016-02-10
Add option to collapse identical columns in HTML report


Athari:
Some target platforms are very similar and analysis can produce identical results for them. Viewing HTML report will be easier if such platforms are collapsed together. Also, when porting, it's helpful to see what platform-dependent code should be moved/replaced/whatever. Overall, the report will be considerably less scary if it contains less redundant information.

twsouthwick:
This makes sense. We would love a contribution if you want to add that to the HTML report generator!

conniey:
Nice to have, please re-open if you have a PR to submit. :)

conniey:
There are checkboxes so you can hide columns (under "Hide columns:") if they are cumbersome to look at. I hope this fixes your issue.

![image](https://cloud.githubusercontent.com/assets/10136526/26808940/888f8362-4a15-11e7-8395-f7f20bb9cfb6.png)

PR #378, opened 2016-11-19
Allow hiding rows that are all green

jeremymeng:
This allows users to focus on Api's that have possible problems in the
current view.

### DIFF
--- a/src/Microsoft.Fx.Portability.Reports.Html/Resources/_Scripts.cshtml
+++ b/src/Microsoft.Fx.Portability.Reports.Html/Resources/_Scripts.cshtml
@@ -67,6 +67,57 @@
         var col_id = parseInt(element.id.substr(-1));
         var visible = !element.checked;
         show_hide_column(col_id, visible);
+        // Refresh the rows because showing columns could make error cells visible.
+        ToggleRowVisibility(document.getElementById('row_visibility_checkbox'));
     }
 
+
+    function ToggleRowVisibility(element){
+        var stl;
+        if (!element.checked) stl = '';
+        else stl = 'none';
+
+        var tables = document.getElementsByClassName('table_portability_details');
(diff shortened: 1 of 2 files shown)


jeremymeng:
/cc @twsouthwick @conniey 

After hiding columns that have the most errors, a lot of the rows now show all green.  Hiding them allows users to focus on useful things.

src/Microsoft.Fx.Portability.Reports.Html/Resources/_Scripts.cshtml
```diff
@@ -67,6 +67,57 @@
...
+                            var cell = cels[cel];
+                            // Keep rows that have visible errors
+                            if (cell.classList.contains('IconErrorEncoded') && cell.style.display != 'none') {
+                                all_cells_success = false;
+                                type_has_visible_child = true;
+                                break;
+                            }
+                        }
+                        // Hide all success row
+                        if (all_cells_success) rows[j].style.display = stl;
```
conniey: I noticed that the code doesn't follow some of the coding styles we have. Can you update that?
[Coding Style](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md)

For example:
* New line for {
* Naming convention: all_cells_success versus allCellsSuccess

jeremymeng: I've keep the "{" since it's the javascript convention, and consistent with existing functions above in the same file.

conniey: Makes sense. I guess I am used to the coding style represented here for JS. https://google.github.io/styleguide/jsguide.html#naming-rules-common-to-all-identifiers

Any thoughts @twsouthwick

conniey:
#261 Helps address the issue by hiding some information.

PR #396, opened 2017-02-09
Consider proxy when connecting to service

conniey:
* Adding support for proxies since we do not consider them.  This feature is modeled after the architecture NuGet uses where it tries to find a proxy in the following order:

    1. Configuration file (config.json)
    2. Environment variable ([http_proxy](https://msdn.microsoft.com/en-us/library/hh272656(v=vs.120).aspx))
    3. System configured

    If none exists, it behaves normally.

* If a 407 is encountered, it has a ProxyAuthenticationHandler to try and fetch the NetworkCredentials from the user.
* Updates IProgressReporter so that asking users for credentials does not conflict with the [Loading...] animations.

Example usage below:

example

__TODOs__

* Update HELP command
* Add documentation
* Add Unit Tests

I am not sure if I should move these tasks to a new PR or keep adding to this one.

### DIFF
--- a/src/ApiPort/ConsoleProgressReporter.cs
+++ b/src/ApiPort/ConsoleProgressReporter.cs
@@ -5,7 +5,6 @@ using ApiPort.Resources;
 using Microsoft.Fx.Portability;
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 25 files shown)


src/ApiPort/ConsoleProgressReporter.cs
```diff
@@ -32,6 +38,28 @@ public void ReportIssue(string issue)
             _issuesReported.Add(issue);
         }
 
+        /// <summary>
+        /// Suspends all progress tasks.
+        /// </summary>
+        public void SuspendProgressTasks()
```
twsouthwick: Suspend/Resume might be better names. Is there value in suffixing with ProgressTasks?

src/ApiPort/ConsoleProgressReporter.cs
```diff
@@ -83,35 +131,53 @@ public ConsoleProgressTask(string task, int? totalCount)
                 await Task.Delay(1);
 
 #if FEATURE_RICH_CONSOLE
```
twsouthwick: I saw you removed this define - do we still need this code?

src/ApiPort/ConsoleProgressReporter.cs
conniey: Good call. I get a little over-wordy with my variable/method names.

src/ApiPort/ConsoleProgressReporter.cs
conniey: I don't think we need this define anymore... it comes in .NET Core too. I'll remove it.

twsouthwick:
LGTM - @conniey is this ready to merge?

conniey:
Yes! I was waiting on the CI builds.

src/ApiPort/Proxy/ProxyProvider.cs
```diff
@@ -0,0 +1,282 @@
...
+                            password: credentials[1]
+                        );
+                    }
+                }
+
+                var noProxy = Environment.GetEnvironmentVariable(ProxyConstants.NoProxy);
+                if (!string.IsNullOrEmpty(noProxy))
+                {
+                    // split comma-separated list of domains
+                    webProxy.BypassList = noProxy.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
```
abatkin: You might consider trimming whitespace from each entry too

PR #403, opened 2017-03-20
Serialize object to Stream rather than string

conniey:
Fix OutOfMemoryExceptions because of serialization to string.

### DIFF
--- a/src/Microsoft.Fx.Portability/DataExtensions.cs
+++ b/src/Microsoft.Fx.Portability/DataExtensions.cs
@@ -38,10 +38,33 @@ namespace Microsoft.Fx.Portability
                 using (var writer = new StreamWriter(outputStream))
                 using (var jsonWriter = new JsonTextWriter(writer))
                 {
-                    Serializer.Serialize(jsonWriter, data);
+                    var serializer = JsonSerializer.Create(JsonSettings);
+                    serializer.Formatting = Formatting.None;
+                    serializer.Serialize(jsonWriter, data);
                 }
 
                 return outputStream.ToArray();
+            }
+        }
+
+        public static byte[] SerializeAndCompress<T>(this T data)
+        {
+            using (var jsonSerializedStream = new MemoryStream())
+            using (var writer = new StreamWriter(jsonSerializedStream))
+            using (var jsonWriter = new JsonTextWriter(writer))
+            {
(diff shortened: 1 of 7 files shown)


src/Microsoft.Fx.Portability/DataExtensions.cs
```diff
@@ -20,7 +20,7 @@ public static class DataExtensions
 
         public static JsonSerializerSettings JsonSettings { get; } = new JsonSerializerSettings
         {
-            Formatting = Formatting.Indented,
+            Formatting = Formatting.None,
```
twsouthwick: The indented is nice for the times when we want to look at it. I'm fine with Formatting.None, but I'm wondering what the size difference is with this especially when the stream gets compressed.

src/Microsoft.Fx.Portability/DataExtensions.cs
```diff
@@ -33,13 +33,13 @@ public static class DataExtensions
...
         public static byte[] Serialize<T>(this T data)
         {
-            var str = JsonConvert.SerializeObject(data, Formatting.Indented, JsonSettings);
-
             using (var outputStream = new MemoryStream())
-            using (var writer = new StreamWriter(outputStream))
             {
-                writer.Write(str);
-                writer.Flush();
+                using (var writer = new StreamWriter(outputStream))
```
twsouthwick: The analyze request also gets compressed. We should rework these helpers so that it keeps items as streams as long as possible. The `.ToArray` creates an intermediary array that may not be needed and would just be excess allocations.

src/Microsoft.Fx.Portability/DataExtensions.cs
conniey: The size difference is about 348KB. I reverted it

src/Microsoft.Fx.Portability/DataExtensions.cs
conniey: I added [`SerializeAndCompress<T>`(this T data)](https://github.com/Microsoft/dotnet-apiport/pull/403/files#diff-e913c19f08e1e43f4dc9847354ec853bR50).  Does this work?

twsouthwick: Yup. We can remove it later

PR #410, opened 2017-04-04
Add a null check for Id in BreakingChange.GetHashCode

twsouthwick:


twsouthwick:
I found another null check... I'll fix that and add some tests

conniey:
Weird. I can't get the NullReferenceException to reproduce when debugging the server.

src/Microsoft.Fx.Portability/BreakingChange.cs
```diff
@@ -87,7 +87,7 @@ public int CompareTo(BreakingChange other)
 
         public override int GetHashCode()
         {
-            return Id.GetHashCode();
+            return Id?.GetHashCode() ?? 0;
```
conniey: If the Id is null, is it better to use `base.GetHashCode()` or some thing else other than 0? In the case that there are multiple BreakingChanges where the Id is not set.

twsouthwick: GetHashCode isn't the same as equality, so I don't think it matters

twsouthwick:
I've fixed the issue in the docs with the null check - remove the colon from one of the entries and you'll see it show up. The parser can't determine the Id without that

src/Microsoft.Fx.Portability/BreakingChange.cs
conniey: The [documentation](https://msdn.microsoft.com/en-us/library/system.object.gethashcode(v=vs.100).aspx#Remarks) is confusing me. :( I didn't think so either but under **Notes to Implementers**:

>  A hash function is used to quickly generate a number (hash code) that corresponds to the value of an object. Hash functions are usually specific to each Type and, for uniqueness, must use at least one of the instance fields as input.
> 
> A hash function must have the following properties:
> 
> * If two objects compare as equal, the GetHashCode method for each object must return the same value. However, if two objects do not compare as equal, the GetHashCode methods for the two object do not have to return different values.

conniey: Never mind. No need to modify this. :) The [updated documentation](https://msdn.microsoft.com/en-us/library/system.object.gethashcode(v=vs.110).aspx) is more clear. If two objects are equal then GetHashCode should be the same for both objects.

It would be better to modify [BreakingChanges.Equals()](https://github.com/Microsoft/dotnet-apiport/blob/master/src/Microsoft.Fx.Portability/BreakingChange.cs) function.. since it would be possible to return true on two separate BreakingChanges that did not have their ID set.

PR #420, opened 2017-05-03
Fix Vsixs at Runtime

conniey:
* At runtime, there were FileNotFoundExceptions because of app.configs I had updated in #411.  In this PR, I removed them because they are not necessary.  Now the vsix runs on VS2015 and VS2017.
* Fixed a NullReferenceException when disposing of OptionsPage


conniey:
@twsouthwick can you take a quick look at this PR? It was mostly because I would experience runtime changes in VS2015... so I had to downgrade to use 14.0 packages.  And I kept running into FileNotFound exceptions because vsix don't use assembly redirects since it is run inside devenv.exe, so it doesn't use app.config. As a result, I had to write AssemblyRedirectResolver.cs

Issue #390, opened 2016-12-16
Could not find file


ygoe:
Running this on an ASP.NET Core 1.0 project targeting .NET Framework 4.6, with the then-proposed directory layout. Here's the output:

```
Microsoft (R) .NET Portability Analyzer 1.0.0.0 (alpha)
Copyright (C) Microsoft Corporation. All rights reserved.
More information is available at http://go.microsoft.com/fwlink/?LinkId=506955

Detecting assembly references                      [Done]
* Could not find file: 'D:\Projekte\MyProjectName\Code\artifacts\bin\MyWebAssemblyName\'
* No files were found to upload.

Finding source line information
	                                               [Failed]
```

bernd5:
This error seem to occur when we use the new csproj file format, really bad.
I have my code in shared projects. If add a reference of it to a "classic" project it works fine, but with the new format it fails...

ygoe:
My original report was for project.json, which was available at that time. But it doesn't seem this tool gets any maintenance. With the upcoming .NET Standard/Core 2.0, this probably isn't much needed anymore anyway.

dkent600:
I'm getting the same error, the analysis is unable to find folders that do exist.  :-(

```
Detecting assembly references                      [Done]
* Could not find file: 'C:\Users\[me]\Documents\Projects\CommonUtilitiesCore\CommonUtilitiesCore\bin\Debug\netcoreapp1.1\'
* Could not find file: 'C:\Users\[me]\Documents\Projects\CommonUtilitiesCore\CommonUtilitiesCoreTest\bin\Debug\netcoreapp1.1\'
* No files were found to upload.


Finding source line information
                                                       [Failed]
                                                       [Failed]
```

conniey:
Hi,

Thanks for the issue feedback. It was because we were looking for `<OutputFileName>` in our ProjectAnalyzer for the binaries output from VS projects. I have a PR open for it #430 but am waiting to merge it in because the solution only works on VS2017 right now.  If you want, you can pull from that branch and build it.

dkent600:
cool, thanks for the update!

PR #430, opened 2017-05-24
Fix locating project output paths

### LINKED ISSUE #397: Add Visual Studio 2017 Support
The VS extension is currently not available for VS 2017. We should add that.

### LINKED ISSUE #429: "Could not find file: '<projectpath>\bin\Debug\net46\' from VSIX
I have a simple project that I have built and compiled in 4.6. I have confirmed that the complied .dll is available in <projectpath>\bin\Debug\net46\ directory . 

Downloaded and installed vsix for portability tool, in Visual Studio 2017.  When I attempt to run the addin I always get the following error with no other information provided.

```
Microsoft (R) .NET Portability Analyzer 1.0.0.0 (alpha)
Copyright (C) Microsoft Corporation. All rights reserved.
More information is available at http://go.microsoft.com/fwlink/?LinkId=506955

Detecting assembly references                      [Done]
* Could not find file: 'C:\src\acme\tfs\Core\Acme.Core\bin\Debug\net46\'
* No files were found to upload.

Finding source line information
                                                                    [Failed]
```

conniey:
* Fixes output path determination for projects. Before we would look for at a project for `<OutputFileName>` and stop if that property could not be determined.
* Supports .NET Core projects now. (Need to verify for project.json projects in VS2015.)
  * **Caveat**:  Released version Microsoft.Cci does not support portable PDB file structure, so it throws an OOM. As a result, I had to add this output window as a workaround... for now.  The repository has support for portable PDB, but I have no idea where the latest builds are. There is an issue here to get the builds: Microsoft/cci#46.
image
* Supports showing the proper projects for source line information. (Before, it would show the first project for all the source line information.)
image
* Fixes UI hangs due to async/await.  As a result, I used the ThreadHelper from this blog and ConfigureAwait(false): https://blogs.msdn.microsoft.com/andrewarnottms/2014/05/07/asynchronous-and-multithreaded-programming-within-vs-using-the-joinabletaskfactory/

Possibly fixes #429
Fixes #397, #390 

/cc: @twsouthwick

### DIFF
--- a/src/ApiPort.VisualStudio/Analyze/ProjectAnalyzer.cs
+++ b/src/ApiPort.VisualStudio/Analyze/ProjectAnalyzer.cs
@@ -7,8 +7,10 @@ using EnvDTE;
 using Microsoft.Fx.Portability;
 using Microsoft.Fx.Portability.Reporting;
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace ApiPortVS.Analyze
@@ -38,9 +40,9 @@ namespace ApiPortVS.Analyze
             _errorList = errorList;
         }
 
-        public async Task AnalyzeProjectAsync(ICollection<Project> projects)
+        public async Task AnalyzeProjectAsync(ICollection<Project> projects, CancellationToken cancellationToken = default(CancellationToken))
         {
-            var buildSucceeded = await _builder.BuildAsync(projects);
+            var buildSucceeded = await _builder.BuildAsync(projects).ConfigureAwait(false);
 
             if (!buildSucceeded)
             {
@@ -48,7 +50,28 @@ namespace ApiPortVS.Analyze
             }
 
             // TODO: Add option to include everything in output, not just build artifacts
-            var targetAssemblies = projects.SelectMany(p => p.GetAssemblyPaths().Where(x => !string.IsNullOrEmpty(x))).ToList();
+            var targetAssemblies = new ConcurrentBag<string>();
+
+            foreach (var project in projects)
+            {
+                var output = await project.GetBuildOutputFilesAsync();
+
+                // Could not find any output files for this. Skip it.
+                if (output == null)
+                {
+                    continue;
+                }
+
+                foreach (var file in output)
+                {
+                    targetAssemblies.Add(file);
+                }
+            }
+
+            if (!targetAssemblies.Any())
+            {
+                throw new PortabilityAnalyzerException(LocalizedStrings.FailedToLocateBuildOutputDir);
+            }
 
             var result = await _analyzer.WriteAnalysisReportsAsync(targetAssemblies, _reportWriter, true);
 
--- a/src/ApiPort.VisualStudio/Analyze/ProjectAnalyzer.cs
+++ b/src/ApiPort.VisualStudio/Analyze/ProjectAnalyzer.cs
@@ -77,7 +77,7 @@ namespace ApiPortVS.Analyze
 
             var sourceItems = await Task.Run(() => _sourceLineMapper.GetSourceInfo(targetAssemblies, result));
 
-            DisplaySourceItemsInErrorList(sourceItems, projects);
+            await DisplaySourceItemsInErrorList(sourceItems, projects);
         }
 
         public bool FileHasAnalyzableExtension(string fileName)
@@ -102,12 +102,14 @@ namespace ApiPortVS.Analyze
             }
         }
 
-        private void DisplaySourceItemsInErrorList(IEnumerable<ISourceMappedItem> items, ICollection<Project> projects)
+        private async Task DisplaySourceItemsInErrorList(IEnumerable<ISourceMappedItem> items, ICollection<Project> projects)
         {
             if (!items.Any())
             {
                 return;
             }
+
+            await Microsoft.VisualStudio.Shell.ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
 
             _errorList.Tasks.Clear();
             _errorList.Refresh();
--- a/src/ApiPort.VisualStudio/Analyze/ProjectAnalyzer.cs
+++ b/src/ApiPort.VisualStudio/Analyze/ProjectAnalyzer.cs
@@ -6,12 +6,15 @@ using ApiPortVS.Resources;
 using EnvDTE;
 using Microsoft.Fx.Portability;
 using Microsoft.Fx.Portability.Reporting;
+using Microsoft.VisualStudio.Shell.Interop;
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+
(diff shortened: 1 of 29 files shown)


src/ApiPort.VisualStudio/Analyze/ProjectAnalyzer.cs
```diff
@@ -38,23 +43,44 @@ public class ProjectAnalyzer
...
                 throw new PortabilityAnalyzerException(LocalizedStrings.UnableToBuildProject);
             }
 
             // TODO: Add option to include everything in output, not just build artifacts
-            var targetAssemblies = projects.SelectMany(p => p.GetAssemblyPaths()).ToList();
+            var targetAssemblies = new ConcurrentBag<string>();
+
+            foreach (var project in projects)
+            {
+                var output = await project.GetBuildOutputFilesAsync();
```
twsouthwick: `ConfigureAwait`? There's an analyzer we could add that will check for these

src/ApiPort.VisualStudio/Analyze/ProjectAnalyzer.cs
```diff
@@ -79,21 +105,43 @@ private IEnumerable<string> GetAllAssembliesInDirectory(string directory)
...
+
+                var hierarchy = project.GetHierarchy();
+
+                foreach (var output in outputs)
+                {
+                    projectWithOutputMappings.AddOrUpdate(output, hierarchy, (existingKey, existingValue) => hierarchy);
+                }
+            }
+
+            await VisualStudio.ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
```
twsouthwick: Do we need this again? I would expect us to still be on the main thread due to 115. The only await we have here doesn't have `ConfigureAwait` so it should be on the same context

src/ApiPort.VisualStudio/ApiPortVSPackage.cs
```diff
@@ -68,7 +68,7 @@ protected override void Initialize()
                 mcs.AddCommand(analyzeMenuOptionsItem);
 
                 CommandID analyzeMenuToolbarCommandID = new CommandID(Guids.AnalyzeMenuItemCmdSet, (int)PkgCmdIDList.CmdIdAnalyzeToolbarMenuItem);
-                MenuCommand analyzeMenuToolbarItem = new MenuCommand((_, __) => ShowToolbar(), analyzeMenuToolbarCommandID);
+                MenuCommand analyzeMenuToolbarItem = new MenuCommand(async (_, __) => await ShowToolbarAsync(), analyzeMenuToolbarCommandID);
```
twsouthwick: Does the `ShowToolbarAsync` need to be `async Task`? If it is only used as an event handler, you can just use `async void`

src/ApiPort.VisualStudio/ApiPortVSPackage.cs
```diff
@@ -100,8 +100,12 @@ protected override void Initialize()
             }
         }
 
-        public void ShowToolbar()
+        public async System.Threading.Tasks.Task ShowToolbarAsync()
         {
+            // Calls to UI elements should use the main task thread.
+            // https://blogs.msdn.microsoft.com/andrewarnottms/2014/05/07/asynchronous-and-multithreaded-programming-within-vs-using-the-joinabletaskfactory/
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
```
twsouthwick: Not sure if VS does stuff differently, but UI events are generally called on the UI thread (at least in WPF/UWP)

src/ApiPort.VisualStudio/DteProjectExtensions.cs
```diff
@@ -1,74 +1,189 @@
...
+        /// Tries to fetch output items if it uses Common Project System then
+        /// tries to fetch output items by retrieving FinalBuildOutput
+        /// location using code snippet from:
+        /// https://github.com/Microsoft/visualfsharp/blob/master/vsintegration/tests/unittests/Tests.ProjectSystem.Miscellaneous.fs#L168-L182
+        /// </summary>
+        /// <returns>null if it is unable to retrieve VS configuration objects</returns>
+        public static async Task<IEnumerable<string>> GetBuildOutputFilesAsync(this Project project, CancellationToken cancellationToken = default(CancellationToken))
         {
-            var fileName = project.Properties.Item("OutputFileName").Value.ToString();
+            var output = await project?.GetBuildOutputFilesFromCPSAsync(cancellationToken);
```
twsouthwick: using the `?.` operator with `await` doesn't work - you'd end up with `await null;` which is not handled. Therefore, you'll need to explicitly check for null before calling this now.

src/ApiPort.VisualStudio/DteProjectExtensions.cs
```diff
@@ -1,74 +1,189 @@
...
+                return output;
+            }
+
+            var configuration = GetVsProjectConfiguration(project);
+
+            if (configuration == null)
+            {
+                return null;
+            }
+            if (!(configuration is IVsProjectCfg2))
```
twsouthwick: This could be rewritten with pattern matching; ie `if(configuration is IVsProjectCfg2 configuration2)`. You'll have to swap the branching here, though

src/ApiPort.VisualStudio/Reporting/VsBrowserReportViewer.cs
```diff
@@ -25,13 +26,13 @@ public VsBrowserReportViewer(IFileSystem fileSystem, TextWriter output, IVsWebBr
...
 
-        public void View(IEnumerable<string> urls)
+        public async Task ViewAsync(IEnumerable<string> urls)
         {
             foreach (var url in urls)
             {
                 if (IsHtml(url))
                 {
-                    ShowHtml(url);
+                    await ShowHtmlAsync(url);
```
twsouthwick: `ConfigureAwait`?

src/ApiPort.VisualStudio/ApiPortVSPackage.cs
conniey: This method is called from the ProjectAnalyzer after analysis, which may not be on the main thread... I wanted to ensure we were always on it.

src/ApiPort.VisualStudio/ApiPortVSPackage.cs
conniey: Its used in the ProjectAnalyzer, so I didn't want to fire and forget about it.

src/ApiPort.VisualStudio/Analyze/ProjectAnalyzer.cs
conniey: I added the analyzer and did ConfigureAwait(false) on everything, so its possible we wouldn't resume on the UI thread.

src/ApiPort.VisualStudio/DteProjectExtensions.cs
conniey: 👍 I totally forgot about this feature.

twsouthwick:
LGTM

conniey:
@twsouthwick Should I merge this PR? I know it'll break VS2015 but will work for VS2017. :(

Issue #359, opened 2016-10-13
VSIX creating folder in default location even though directory settings have changed


conniey:
Whenever I open the Visual Studio options, the extension is creating a "Portability Analysis" folder in the "My Documents" directory which is the default output directory that you can specify under options for the extension even though I asked it to use a different directory for the output.

[Original comment](https://github.com/Microsoft/dotnet-apiport/issues/355#issuecomment-253633450)

PR #438, opened 2017-06-06
Fix crash when opening report in VS

### LINKED ISSUE #402: Attempting to open an analysis report that has been deleted crashes VS
**Repro steps:**
* analyse a dll using the VS extension
* click on _Open directory_ in the _Portability Analysis Results_ tool window and delete/rename the report file
* click on _Open Report_ in the _Portability Analysis Results_

**Expected:**
Anything except a crash (e.g. a message box saying the file has been moved or deleted)

**Actual:**
VS crashes

twsouthwick:
This will check if the report (or directory) is available and if it is, will launch it. Otherwise, a notification is given and then the entries are removed from the pane. This change also moves to create the directory to output to if it is not already there

Fixes #402 and #359

### DIFF
--- a/src/ApiPort.VisualStudio/Models/OptionsModel.cs
+++ b/src/ApiPort.VisualStudio/Models/OptionsModel.cs
@@ -43,7 +43,7 @@ namespace ApiPortVS.Models
             LastUpdate = DateTimeOffset.MinValue;
             OutputDirectory = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments), "Portability Analysis");
         }
-        
+
         public DateTimeOffset LastUpdate { get; set; }
 
         public IList<SelectedResultFormat> Formats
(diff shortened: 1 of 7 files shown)


src/ApiPort.VisualStudio/Models/OptionsModel.cs
```diff
@@ -61,24 +61,13 @@ public IList<TargetPlatform> Platforms
...
-                    Directory.CreateDirectory(_outputDirectory);
-                }
-            }
+            set { UpdateProperty(ref _outputDirectory, value); }
         }
 
         public string DefaultOutputName
         {
             get { return _defaultOutputName; }
-            set
```
conniey: Is it possible that they'll set the DefaultOutputName to "" or null?

src/ApiPort.VisualStudio/Models/OptionsModel.cs
```diff
@@ -61,24 +61,13 @@ public IList<TargetPlatform> Platforms
         public string OutputDirectory
         {
             get { return _outputDirectory; }
-            set
-            {
-                UpdateProperty(ref _outputDirectory, value);
```
conniey: The default SelectedPath for the dialog is "", so I was wondering if it was possible to press OK and not select a folder on the dialog.

src/ApiPort.VisualStudio/ViewModels/OutputViewModel.cs
```diff
@@ -71,9 +103,7 @@ public DelegateCommand(Action<T> action)
 
             public void Execute(object parameter)
             {
-                var obj = parameter as T;
-
-                if (obj != null)
+                if (parameter is T obj)
```
conniey: 👍

src/ApiPort.VisualStudio/Models/OptionsModel.cs
twsouthwick: Good point - I've updated this so it will check for null or whitespace and default to a reasonable place

PR #453, opened 2017-07-19
Remove instance variable to track System.Object assembly

twsouthwick:
A recent change (#452) went in that added an instance variable to the DependencyFinderEngineHelper that tracks the assembly System.Object is in. However, this instance variable introduces state that is unnecessary and that makes the routine non-thread safe.

Also, there was a check to only calculate this if the assembly filter is of a particular form - this removes that so we always find the assembly.

### DIFF
--- a/src/Microsoft.Fx.Portability.MetadataReader/DependencyFinderEngineHelper.cs
+++ b/src/Microsoft.Fx.Portability.MetadataReader/DependencyFinderEngineHelper.cs
@@ -71,13 +71,8 @@ namespace Microsoft.Fx.Portability.Analyzer
             }
 
             // Get the assembly info of System.Object and set it as assembly info for primitives
-            var systemObjectMemberDependency = MemberDependency.FirstOrDefault(t => string.Equals(t.MemberDocId, "T:System.Object", StringComparison.Ordinal) && _assemblyFilter.IsFrameworkAssembly(t.DefinedInAssemblyIdentity));
+            var systemObjectMemberDependency = MemberDependency.FirstOrDefault(t => string.Equals(t.MemberDocId, "T:System.Object", StringComparison.Ordinal));
             var systemObjectAssembly = systemObjectMemberDependency?.DefinedInAssemblyIdentity;
-
-            if (systemObjectAssembly == null)
-            {
-                throw new PortabilityAnalyzerException(LocalizedStrings.MissingAssemblyInfo);
-            }
 
             // Get member references
             foreach (var handle in _reader.MemberReferences)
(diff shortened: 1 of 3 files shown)


src/Microsoft.Fx.Portability.MetadataReader/DependencyFinderEngineHelper.cs
```diff
@@ -73,27 +71,25 @@ public void ComputeData()
             }
 
             // Get the assembly info of System.Object and set it as assembly info for primitives
-            if (_assemblyFilter is DotNetFrameworkFilter)
+            var systemObjectMemberDependency = MemberDependency.FirstOrDefault(t => string.Equals(t.MemberDocId, "T:System.Object", StringComparison.Ordinal) && _assemblyFilter.IsFrameworkAssembly(t.DefinedInAssemblyIdentity));
```
twsouthwick: Why do you need to check if it is a framework assembly? Theoretically, you could create your own framework with System.Object somewhere else not in the framework. I'm going to remove this unless someone has a reason why we need to filter it.

conniey: Hmm, that is true. If they created their own framework, would they want to check for API compatibility with our .NET framework APIs? We use [DotNetFrameworkFilter](https://github.com/Microsoft/dotnet-apiport/blob/a5988f7c0cd01ce31b83ea7bdee39b8abf4d4c38/src/Microsoft.Fx.Portability/Analyzer/DotNetFrameworkFilter.cs) to calculate which APIs to send up. I would opt to keep for the consistency. but I am OK either way.

twsouthwick: I've seen a customer in precisely that situation :) Since the filters are meant more as to calculate which ones should be included, I'm going to remove it

alinapopa: But without this check there is the possibility of being another System.Object defined somewhere else.

twsouthwick: Right - that was my initial comment because you can do that. However, if you are creating your own System.Object that is supposed to hide the system System.Object, we will miss that...

alinapopa: Without these two checks there can be the following undesired situations which are not addressed:

- there is more than one System.Object defined
- there is no System.Object found (see the failing test)
- there is one System.Object found but it is something defined in customer's assembly

twsouthwick: Ok - I'll add the check back.

PR #467, opened 2017-09-01
Add excel writer

twsouthwick:
Now that DocumentForma.OpenXml is availalbe, we can put our excel writer here.


twsouthwick:
@conniey I've got a PR that will simplify the packaging (included in #475), so I'd rather wait to incorporate the package creation there. If that's okay, once it's merged, I'll update that PR

twsouthwick:
@conniey I've rebased against my changes for PackageReference so now the nupkg is generated for this library

PR #468, opened 2017-09-06
Fixes NullReferenceException when seeing data being transmitted

conniey:
* When using [See the data being transmitted](https://github.com/Microsoft/dotnet-apiport/tree/master/docs/Console#see-the-data-being-transmitted), if format is not specified, it throws a NullReferenceException during analysis.

There is only one supported format, Json, this returns that.


conniey:
@alinapopa Can you check this PR to make sure it is OK?

PR #482, opened 2017-09-18
Remove all but one binding redirection compile warning

twsouthwick:


src/Microsoft.Fx.Portability/Microsoft.Fx.Portability.csproj
```diff
@@ -14,25 +14,22 @@
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
     <PackageReference Include="Microsoft.AspNetCore.WebUtilities" Version="1.1.1" />
     <PackageReference Include="System.Net.Http" Version="4.3.2" />
+    <PackageReference Include="System.Collections.Immutable" Version="1.3.1" />
```
conniey: I remember @alinapopa  moving this into the netstandard1.3 and net46 TargetFrameworks because at runtime, it would fail to resolve the System.Collections.Immutable dependency.  Does it still work in both VS2017 and VS2015?

conniey:
Wow! Awesome! What was the thing that fixed all these?  I noticed the migration to .NET Core project style and `<PrivateAssets>All</PrivateAssets>`? Was there something else?

src/Microsoft.Fx.Portability/Microsoft.Fx.Portability.csproj
twsouthwick: Yup - the issue was that we hadn't been including System.Collections.Immutable with the VSIX. It's strongly named and we're not sharing it with other systems (such as Roslyn) so we should be okay. It works in all cases

twsouthwick:
That was mostly it - there were also issues introduced from different versions of some libraries (such as System.Collections.Immutable)

src/Microsoft.Fx.Portability/Microsoft.Fx.Portability.csproj
alinapopa: In this case is the net46 target still needed?

twsouthwick: probably not - I won't remove that target right now so this change won't be blocked, but we can look into it.

PR #483, opened 2017-09-18
Add offline project file

twsouthwick:
With the simplified project structure the new csproj format allows, we can easily create another project file that generates an ApiPort.Offline version.


src/ApiPort/ApiPort.Offline.csproj
```diff
@@ -0,0 +1,65 @@
...
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net46'">
+    <RuntimeIdentifier>win7-x64</RuntimeIdentifier>
+    <DefineConstants>$(DefineConstants);FEATURE_SYSTEM_PROXY;FEATURE_NETWORK_CREDENTIAL</DefineConstants>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <!-- Projects default to being signed, but Microsoft.Configuration.* assemblies are not signed -->
```
conniey: Maybe in a unrelated PR.. but Microsoft.Extensions.Configuration is signed now... Should we remove this?

twsouthwick: Yeah - but we should probably switch to Microsoft.CommandLine from https://github.com/dotnet/corefxlab

PR #485, opened 2017-09-19
Add FxCop analyzers and fix up culture aware formatting

twsouthwick:
The official builds use FxCop but we don't have anything at the moment to help the dev experience. Right now, many analyzers are disabled (look at `rule.ruleset`), but we can look at enabling them in the future. This sets the analyzers to catch the items that the official build system uses.


twsouthwick:
There seems to be an issue loading right now from CI - closing for now

PR #486, opened 2017-09-19
Add culture aware formatting

twsouthwick:
FxCop is failing in the official build, so adding the fixes here

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -5,7 +5,7 @@
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
     <OutputDrop>$(TF_BUILD_BINARIESDIRECTORY)</OutputDrop>
-    <NoWarn>$(NoWarn),1570,1572,1573,1574,1591,1701,CA0068</NoWarn>
+    <NoWarn>$(NoWarn),1570,1572,1573,1574,1591,1701</NoWarn>
   </PropertyGroup>
 
   <PropertyGroup>
@@ -22,6 +22,14 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)\PortabilityTools.snk</AssemblyOriginatorKeyFile>
     <DefineConstants>$(DefineConstants);FEATURE_STRONGNAMESIGNING</DefineConstants>
+  </PropertyGroup>
+
+  <!-- FxCop does not understand this target platform and will output an error similar to the following:
+    MSBUILD : error : CA0055 : Could not identify platform for 'bin\Debug\ApiPort\netcoreapp1.0\ApiPort.dll'
+    MSBUILD : error : CA0052 : No targets were selected.
+    -->
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp1.0'">
+    <RunCodeAnalysis>false</RunCodeAnalysis>
   </PropertyGroup>
 
   <!-- Do not generate AssemblyInfo because we are doing that with the
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -5,7 +5,7 @@
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
     <OutputDrop>$(TF_BUILD_BINARIESDIRECTORY)</OutputDrop>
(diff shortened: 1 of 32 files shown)


src/Microsoft.Fx.Portability.Reports.Excel/ExcelOpenXmlOutputWriter.cs
```diff
@@ -152,7 +152,7 @@ private void GenerateSummaryPage(Worksheet summaryPage, ReportingResult analysis
                     var summaryData = new List<object>() { analysisResult.GetNameForAssemblyInfo(item.SourceAssembly), item.SourceAssembly.TargetFrameworkMoniker ?? string.Empty };
 
                     // TODO: figure out how to add formatting to cells to show percentages.
-                    summaryData.AddRange(item.UsageData.Select(pui => (object)((pui.PortabilityIndex == 0) ? 0 : double.Parse((pui.PortabilityIndex * 100).ToString("##.##")))));
+                    summaryData.AddRange(item.UsageData.Select(pui => (object)(Math.Round(pui.PortabilityIndex * 100.0, 2))));
```
twsouthwick: Not sure why it was formatted out to two decimals and then parsed back - I think rounding will do the same thing

Directory.Build.props
```diff
@@ -5,7 +5,7 @@
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
     <OutputDrop>$(TF_BUILD_BINARIESDIRECTORY)</OutputDrop>
-    <NoWarn>$(NoWarn),1570,1572,1573,1574,1591,1701</NoWarn>
+    <NoWarn>$(NoWarn),1570,1572,1573,1574,1591,1701,CA0068</NoWarn>
```
conniey: Do you know why we get this warning? I see .pdbs next to all of our assemblies...

src/ApiPort/Proxy/ProxyProvider.cs
```diff
@@ -243,7 +243,7 @@ private static bool IsSystemProxySet(Uri uri)
                 if (proxyUri != null)
                 {
                     var proxyAddress = new Uri(proxyUri.AbsoluteUri);
-                    if (string.Equals(proxyAddress.AbsoluteUri, uri.AbsoluteUri))
+                    if (string.Equals(proxyAddress.AbsoluteUri, uri.AbsoluteUri, StringComparison.Ordinal))
```
conniey: Hmm... should it be `StringComparison.OrdinalIgnoreCase`?
https://stackoverflow.com/questions/15641694/are-uris-case-insensitive

tests/Microsoft.Fx.Portability.Tests/ApiPortServiceTests.cs
```diff
@@ -12,9 +12,9 @@
 
 namespace Microsoft.Fx.Portability.Tests
 {
-    public class ApiPortServiceTests
+    public sealed class ApiPortServiceTests : IDisposable
```
conniey: I'm curious why this class is sealed but the others are not? :)

tests/ApiPort.Tests/ApiPort.Tests.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp1.0</TargetFramework>
+    <RunCodeAnalysis>false</RunCodeAnalysis>
```
conniey: Do we skip CodeAnalysis for all the other test projects too?

tests/Microsoft.Fx.Portability.Tests/ApiPortServiceTests.cs
twsouthwick: This one creates a disposable member field - the sealed is so we don't need to implement the whole dispose pattern

src/ApiPort/Proxy/ProxyProvider.cs
twsouthwick: Good point. I wasn't sure  - I've swapped it to ignore case

Directory.Build.props
twsouthwick: Yeah - probably because we don't create "full" pdbs. FxCop is out of date with the new project/debug types. This suppression doesn't work anyway, so I'll remove it.

tests/ApiPort.Tests/ApiPort.Tests.csproj
twsouthwick: No - just for netcoreapp1.0 types. I've moved the `RunCodeAnalysis` setting to a conditional on `TargetFramework==netcoreapp1.0` in the `Directory.Build.props`

PR #487, opened 2017-09-20
Update build scripts to work on official build server

twsouthwick:


Directory.Build.props
```diff
@@ -45,6 +37,19 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
 
+  <Choose>
```
conniey: I never did like the Choose, When, Otherwise tags.. they seem so verbose (and span multiple lines because you have to open and close all those tags). imho, it seems more succinct to write it as:

```xml
<PropertyGroup>
   <IsTest>false</IsTest>
   <IsTest Condition="$(MSBuildProjectName.ToLowerInvariant().Contains('test'))">true</IsTest>
<PropertyGroup>
```

Directory.Build.targets
```diff
@@ -79,7 +80,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="MicroBuild.Core" Version="0.2.0">
+    <PackageReference Include="MicroBuild.Core" Version="0.2.0" Condition="!$(IsTest)">
```
conniey: Good to know! I always see something like: '$(IsTest)' != 'true'

Directory.Build.props
twsouthwick: I originally had that, but I switched to the choose/when because it seemed cleaner :) I'll switch it back

PR #497, opened 2017-10-06
Use GitVersion for versioning

twsouthwick:


conniey:
@twsouthwick do you know if this will work on Linux machines?  I didn't think it was migrated to work on .NET Core

twsouthwick:
Need to redo this PR against dev

PR #499, opened 2017-10-30
Fixes #479

### LINKED ISSUE #479: An unexpected error was encountered while parsing the metadata for the PE file
> 
> This tool analyzes .NET assemblies to determine possible problems moving between .NET platforms (such as
> Windows Store, desktop, Mono, .NET Core, etc) as well as between .NET Framework versions (ie 4.x->4.y).
> 
> To learn more about how this tool works, including the data we are collecting,
> go here - http://go.microsoft.com/fwlink/?LinkId=397652
> 
> Loading catalog                                   [Done]
> Loading breaking changes                          [Done]
> Detecting assembly references                     [Failed]
> 
> An unexpected error was encountered while parsing the metadata for the PE file located at c:\Foo.dll.
> 
> Cannot locate assembly info for System.Object

conniey:
* Fixes #479 by examining assembly references for mscorlib and System.Runtime assemblies for System.Object.
* Adds tests for SystemObjectNotFoundException

/cc @twsouthwick

### DIFF
--- a/src/Microsoft.Fx.Portability.MetadataReader/DependencyFinderEngineHelper.cs
+++ b/src/Microsoft.Fx.Portability.MetadataReader/DependencyFinderEngineHelper.cs
@@ -16,7 +16,8 @@ namespace Microsoft.Fx.Portability.Analyzer
         private static readonly HashSet<string> s_systemObjectAssemblies = new HashSet<string>(StringComparer.Ordinal)
         {
             "mscorlib",
-            "System.Runtime"
+            "System.Runtime",
+            "System.Private.CoreLib"
         };
 
         private readonly IDependencyFilter _assemblyFilter;
--- a/src/Microsoft.Fx.Portability.MetadataReader/DependencyFinderEngineHelper.cs
+++ b/src/Microsoft.Fx.Portability.MetadataReader/DependencyFinderEngineHelper.cs
@@ -1,10 +1,12 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Microsoft.Fx.Portability.Analyzer.Exceptions;
 using Microsoft.Fx.Portability.ObjectModel;
 using Microsoft.Fx.Portability.Resources;
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Reflection.Metadata;
 
 namespace Microsoft.Fx.Portability.Analyzer
@@ -44,7 +46,11 @@ namespace Microsoft.Fx.Portability.Analyzer
 
         public void ComputeData()
         {
-            AssemblyReferenceInformation systemObjectAssembly = null;
+            // Primitives need to have their assembly set, so we search for a
+            // reference to System.Object that is considered a possible
+            // framework assembly and use that for any primitives that don't
+            // have an assembly
+            var systemObjectAssembly = GetSystemRuntimeAssemblyInformation();
 
             var provider = new MemberMetadataInfoTypeProvider(_reader);
 
@@ -62,16 +68,6 @@ namespace Microsoft.Fx.Portability.Analyzer
                     {
                         MemberDependency.Add(typeReferenceMemberDependency);
                     }
-
-                    // Primitives need to have their assembly set, so we search for a reference to System.Object that is considered a possible framework
-                    // assembly and use that for any primitives that don't have an assembly
-                    if (systemObjectAssembly == null
-                        && string.Equals(typeInfo.Namespace, "System", StringComparison.Ordinal)
-                        && string.Equals(typeInfo.Name, "Object", StringComparison.Ordinal)
-                        && _assemblyFilter.IsFrameworkAssembly(assembly))
-                    {
-                        systemObjectAssembly = assembly;
-                    }
                 }
                 catch (BadImageFormatException)
                 {
@@ -82,11 +78,6 @@ namespace Microsoft.Fx.Portability.Analyzer
                     // we can skip such malformed references and just analyze those
                     // that we can successfully decode.
                 }
-            }
-
-            if (systemObjectAssembly == null)
-            {
-                throw new PortabilityAnalyzerException(LocalizedStrings.MissingAssemblyInfo);
             }
 
             // Get member references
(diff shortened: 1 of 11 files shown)


src/Microsoft.Fx.Portability.MetadataReader/DependencyFinderEngineHelper.cs
```diff
@@ -198,5 +189,38 @@ private string GetPrefix(MemberReference memberReference)
...
+            const string mscorlib = nameof(mscorlib);
+            const string SystemRuntime = "System.Runtime";
+
+            var microsoftAssemblies = _reader.AssemblyReferences
+                .Select(handle =>
+                {
+                    var assembly = _reader.GetAssemblyReference(handle);
+                    return _reader.FormatAssemblyInfo(assembly);
+                })
+                .Where(assembly => _assemblyFilter.IsFrameworkAssembly(assembly));
```
twsouthwick: you could just do `.Where(_assemblyFIlter.IsFrameworkAssembly)`

src/Microsoft.Fx.Portability.MetadataReader/Exceptions/SystemObjectNotFoundException.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Microsoft.Fx.Portability.Analyzer.Exceptions
+{
+    /// <summary>
+    /// Exception thrown when assembly containing <see cref="System.Object"/>
+    /// cannot be found.
+    /// </summary>
+    public class SystemObjectNotFoundException : PortabilityAnalyzerException
+    {
+        public SystemObjectNotFoundException(IEnumerable<AssemblyReferenceInformation> assemblies)
```
twsouthwick: may be better to persist the list of items and override the message property

tests/Microsoft.Fx.Portability.MetadataReader.Tests/ManagedMetadataReaderTests.cs
```diff
@@ -139,7 +126,7 @@ public void VerifyDotNetFrameworkFilter()
             Assert.Equal(expected, foundDocIds);
         }
 
-        [Fact(Skip = "Metadata reader currently requires dependency filters to say System.Object is in a framework assembly")]
+        [Fact]
```
twsouthwick: yay :)

tests/Microsoft.Fx.Portability.MetadataReader.Tests/ManagedMetadataReaderTests.cs
```diff
@@ -227,6 +220,23 @@ public void MultidimensionalPrimitiveArray()
...
+        public void ThrowsSystemObjectNotFoundException()
+        {
+            var dependencyFilter = Substitute.For<IDependencyFilter>();
+            dependencyFilter.IsFrameworkAssembly(Arg.Any<AssemblyReferenceInformation>()).Returns(false);
+            var dependencyFinder = new ReflectionMetadataDependencyFinder(dependencyFilter);
+            var assemblyToTest = TestAssembly.Create("FilterApis.cs", false, new[] { typeof(Image).GetTypeInfo().Assembly.Location });
+            var progressReporter = Substitute.For<IProgressReporter>();
+
+            var exception = Assert.Throws<AggregateException>(() =>
```
twsouthwick: Why is this throwing an aggregate exception? not a blocker, but we should only be throwing a single exception

tests/Microsoft.Fx.Portability.MetadataReader.Tests/ManagedMetadataReaderTests.cs
```diff
@@ -227,6 +220,23 @@ public void MultidimensionalPrimitiveArray()
             Assert.Equal(systemObject.DefinedInAssemblyIdentity, definedInAssemblyIdentity);
         }
 
+        [Fact]
+        public void ThrowsSystemObjectNotFoundException()
+        {
+            var dependencyFilter = Substitute.For<IDependencyFilter>();
+            dependencyFilter.IsFrameworkAssembly(Arg.Any<AssemblyReferenceInformation>()).Returns(false);
+            var dependencyFinder = new ReflectionMetadataDependencyFinder(dependencyFilter);
+            var assemblyToTest = TestAssembly.Create("FilterApis.cs", false, new[] { typeof(Image).GetTypeInfo().Assembly.Location });
```
twsouthwick: Why `System.Drawing.Image`? Doesn't that make this test not portable?

src/Microsoft.Fx.Portability.MetadataReader/DependencyFinderEngineHelper.cs
```diff
@@ -198,5 +189,38 @@ private string GetPrefix(MemberReference memberReference)
...
+
+            var mscorlibAssembly = microsoftAssemblies.SingleOrDefault(x => string.Equals(x.Name, mscorlib, StringComparison.Ordinal));
```
twsouthwick: Would it be better to have a hashset of possible locations for system.object? it would make it easier if we need to exten it in the future.

src/Microsoft.Fx.Portability.MetadataReader/DependencyFinderEngineHelper.cs
conniey: Thanks! I always forget you can do that.

src/Microsoft.Fx.Portability.MetadataReader/Exceptions/SystemObjectNotFoundException.cs
conniey: kk will fix

tests/Microsoft.Fx.Portability.MetadataReader.Tests/ManagedMetadataReaderTests.cs
conniey: True. I was trying to find a type that required an additional assembly reference. I'll fix this

tests/Microsoft.Fx.Portability.MetadataReader.Tests/ManagedMetadataReaderTests.cs
conniey: This is wrapped inside a Parallel.ForEach for because we try to compute multiple assemblies in paralle.

PR #507, opened 2017-11-16
Explicitly require TLS 1.2

mjrousos:
ApiPort already uses TLS 1.2 by default, this just makes that requirement explicit so that we're always sure to use the TLS 1.2 protocol.

The PR is slightly complicated by the fact the NetCore and NetFX have different ways of doing this.

### DIFF
--- a/src/Microsoft.Fx.Portability/CompressedHttpClient.cs
+++ b/src/Microsoft.Fx.Portability/CompressedHttpClient.cs
@@ -21,17 +21,17 @@ namespace Microsoft.Fx.Portability
 {
     internal class CompressedHttpClient : HttpClient
     {
-        internal const SslProtocols SupportedSSLProtocols = SslProtocols.Tls12;
-
 #if FEATURE_SERVICE_POINT_MANAGER
         internal const SecurityProtocolType SupportedSecurityProtocols = SecurityProtocolType.Tls12;
+#else
+        internal const SslProtocols SupportedSSLProtocols = SslProtocols.Tls12;
 #endif
 
         /// <param name="productName">Product name that will be displayed in the User Agent string of requests</param>
         /// <param name="productVersion">Product version that will be displayed in the User Agent string of requests</param>
         public CompressedHttpClient(ProductInformation info)
             : this(info, new HttpClientHandler {
-#if FEATURE_CLIENT_HANDLER_SET_SSL_PROTOCOLS
+#if !FEATURE_SERVICE_POINT_MANAGER
                 SslProtocols = SupportedSSLProtocols, 
 #endif
                 AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate
--- a/src/Microsoft.Fx.Portability/CompressedHttpClient.cs
+++ b/src/Microsoft.Fx.Portability/CompressedHttpClient.cs
@@ -14,22 +14,38 @@ using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Security.Authentication;
 using System.Threading.Tasks;
 
 namespace Microsoft.Fx.Portability
 {
     internal class CompressedHttpClient : HttpClient
     {
+        internal const SslProtocols SupportedSSLProtocols = SslProtocols.Tls12;
+
+#if FEATURE_SERVICE_POINT_MANAGER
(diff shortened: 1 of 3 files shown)


src/Microsoft.Fx.Portability/CompressedHttpClient.cs
```diff
@@ -14,22 +14,38 @@
...
+
+#if FEATURE_SERVICE_POINT_MANAGER
+        internal const SecurityProtocolType SupportedSecurityProtocols = SecurityProtocolType.Tls12;
+#endif
+
         /// <param name="productName">Product name that will be displayed in the User Agent string of requests</param>
         /// <param name="productVersion">Product version that will be displayed in the User Agent string of requests</param>
         public CompressedHttpClient(ProductInformation info)
-            : this(info, new HttpClientHandler { AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate })
+            : this(info, new HttpClientHandler {
```
twsouthwick: Do we need this overload? We may be able to remove it to minimize duplication. We're already injecting a handler in the ApiPortService.

src/Microsoft.Fx.Portability/CompressedHttpClient.cs
```diff
@@ -14,22 +14,38 @@
...
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Security.Authentication;
 using System.Threading.Tasks;
 
 namespace Microsoft.Fx.Portability
 {
     internal class CompressedHttpClient : HttpClient
     {
+        internal const SslProtocols SupportedSSLProtocols = SslProtocols.Tls12;
```
twsouthwick: Is this only used under FEATURE_CLIENT_HANDLER_SET_SSL_PROTOCOLS?

twsouthwick:
It would be nice if the feature flags for this were inverses of each other - otherwise it's not transparent that they have mutually exclusive uses.

src/Microsoft.Fx.Portability/CompressedHttpClient.cs
mjrousos: Yes, I'll move it under an #if to make that clear.

src/Microsoft.Fx.Portability/CompressedHttpClient.cs
mjrousos: Both overloads are used, currently, depending on whether `ApiPortService` is constructed with an `IProxyProvider` or not. The one without the proxy provider is only used by tests and SeachFxApi, so I'm sure the code could be changed to only need one overload but that might belong in another PR.

mjrousos:
Yea, that's a good point. I made two #ifs since it's not the case that there's a "preferred way" and a fallback. There are two ways of doing this and both are unsupported in some .NET profiles. Since we only build for net46 and netstandard1.3, I could change this to just have a single #if (probably FEATURE_SERVICE_POINT_MANAGER) and replace #if FEATURE_CLIENT_HANDLER_SET_SSL_PROTOCOLS with #if !FEATURE_SERVICE_POINT_MANAGER.

src/Microsoft.Fx.Portability/CompressedHttpClient.cs
twsouthwick: Ok. We can remove not a blocker

src/Microsoft.Fx.Portability/Microsoft.Fx.Portability.csproj
```diff
@@ -7,9 +7,9 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TargetFramework)'== 'net46'">
-    <DefineConstants>$(DefineConstants);FEATURE_SERIALIZABLE;FEATURE_ASSEMBLY_LOCATION;FEATURE_XML_SCHEMA</DefineConstants>
+    <DefineConstants>$(DefineConstants);FEATURE_SERIALIZABLE;FEATURE_ASSEMBLY_LOCATION;FEATURE_XML_SCHEMA;FEATURE_SERVICE_POINT_MANAGER</DefineConstants>
   </PropertyGroup>
-
+
```
twsouthwick: nit: extra whitespace?

mjrousos: Good catch. Fixing...

PR #508, opened 2017-11-16
Add StyleCop and standardize whitespace and license

twsouthwick:
This change enables basic style cop analysis, but most of the rules are turned off. The following are turned on: 

- License must be on every file
- Only a single newline between statements
- Newline at end of file
- File must be UTF8


twsouthwick:
@dotnet-bot test this please

PR #514, opened 2017-11-18
Fix breaking change parser

conniey:
Fixes issues in breaking change parser where:
* Comments in suggestions were not being parsed
* Would not trim whitespaces in categories and treat them as different categories.
* Adds tests

### DIFF
--- a/src/Microsoft.Fx.Portability/BreakingChangeParser.cs
+++ b/src/Microsoft.Fx.Portability/BreakingChangeParser.cs
@@ -201,9 +201,12 @@ namespace Microsoft.Fx.Portability
                                 && int.TryParse(contents.Substring(startIndex + 1), out id))
                             {
                                 currentBreak.Id = id.ToString(CultureInfo.InvariantCulture);
+                                state = ParseState.None;
                             }
-
-                            state = ParseState.None;
+                            else
+                            {
+                                ParseNonStateChange(currentBreak, state, currentLine, allowedCategories);
+                            }
                         }
                         // Otherwise, process according to our current state
                         else
@@ -230,7 +233,10 @@ namespace Microsoft.Fx.Portability
                 case ParseState.None:
                     return;
                 case ParseState.OriginalBug:
-                    currentBreak.BugLink = currentLine.Trim();
+                    if (string.IsNullOrEmpty(currentBreak.BugLink))
+                    {
+                        currentBreak.BugLink = currentLine.Trim();
+                    }
                     break;
                 case ParseState.Scope:
                     BreakingChangeImpact scope;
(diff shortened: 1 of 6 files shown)


tests/Microsoft.Fx.Portability.Tests/BreakingChangeParserTests.cs
```diff
@@ -187,7 +241,26 @@ private void TestEquality(BreakingChange expected, BreakingChange actual)
...
+                for (int i = 0; i < resources.Length; i++)
+                {
+                    var resource = resources[i];
+                    _output.WriteLine($"\t{i}: {resource}");
+                }
+
+                throw;
+            }
+
             return typeof(BreakingChangeParserTests).GetTypeInfo().Assembly.GetManifestResourceStream(name);
```
twsouthwick: This can be moved up so that the `name` variable can have its declaration and definition combined

tests/Microsoft.Fx.Portability.Tests/Microsoft.Fx.Portability.Tests.csproj
```diff
@@ -33,6 +33,8 @@
   <ItemGroup>
     <EmbeddedResource Include="TestAssets\005- ListT.ForEach.md" />
     <EmbeddedResource Include="TestAssets\006- System.Uri.md" />
+    <EmbeddedResource Include="TestAssets\CategoryWithSpaces.md" />
```
twsouthwick: It may make it simpler to just replace all these with `<EmbeddedResource Include="TestAssets\*.md" />` or even just a `TestAssets\**\*` glob

src/Microsoft.Fx.Portability/BreakingChangeParser.cs
```diff
@@ -230,7 +233,10 @@ private static void ParseNonStateChange(BreakingChange currentBreak, ParseState
                 case ParseState.None:
                     return;
                 case ParseState.OriginalBug:
-                    currentBreak.BugLink = currentLine.Trim();
+                    if (string.IsNullOrEmpty(currentBreak.BugLink))
+                    {
+                        currentBreak.BugLink = currentLine.Trim();
```
twsouthwick: is this change covered by a test?

src/Microsoft.Fx.Portability/BreakingChangeParser.cs
```diff
@@ -318,7 +324,7 @@ private static void ParseNonStateChange(BreakingChange currentBreak, ParseState
                     {
                         throw new InvalidOperationException(string.Format(CultureInfo.CurrentCulture, LocalizedStrings.InvalidCategoryDetected, category));
                     }
-                    currentBreak.Categories.Add(category);
+                    currentBreak.Categories.Add(category.Trim());
```
twsouthwick: Is this change covered by a test?

src/Microsoft.Fx.Portability/BreakingChangeParser.cs
conniey: Fixed.

tests/Microsoft.Fx.Portability.Tests/BreakingChangeParserTests.cs
conniey: Fixed! :)

src/Microsoft.Fx.Portability/BreakingChangeParser.cs
conniey: Yes. Fixed. :)

PR #521, opened 2017-11-20
Consistent text alignment in REPL

chlowell:
A small fix to ensure `DocIdSearchRepl`'s text alignment remains consistent across localizations.

### DIFF
--- a/src/ApiPort/Resources/LocalizedStrings.resx
+++ b/src/ApiPort/Resources/LocalizedStrings.resx
@@ -341,7 +341,7 @@ Versions marked with an asterisk (*) implies that these are default targets if n
     <value>#count=</value>
   </data>
   <data name="ReplOptionCount_Text" xml:space="preserve">
-    <value>[number]  Display [number] of search results.</value>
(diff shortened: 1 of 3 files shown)


src/ApiPort/Resources/LocalizedStrings.resx
```diff
@@ -341,7 +341,7 @@ Versions marked with an asterisk (*) implies that these are default targets if n
     <value>#count=</value>
   </data>
   <data name="ReplOptionCount_Text" xml:space="preserve">
-    <value>[number]  Display [number] of search results.</value>
+    <value>Display [number] of search results.</value>
```
twsouthwick: Should this be `Display {0} of search results.`?

chlowell: No, it's help text describing the option, never used as a format string. Agree it's awkward. The strings could be refactored such that this one is formatted with the localized "number" string, and perhaps there's a still better way which refactors the string matching in the REPL loop as well. I'd call this a 90% solution we can revisit later should it seem valuable.

PR #531, opened 2017-11-30
Add proxy for visual studio

### LINKED ISSUE #528: Unable to contact service error.
After installing the ,NET Portability Anaylzer extension in VS 2017, I see the error:

Unable to contact service. Try again later.

And hence unable to use the utility. Can someone please shed light on what am I missing?

twsouthwick:
Fixes #528

### DIFF
--- a/src/Microsoft.Fx.Portability/ApiPortService.cs
+++ b/src/Microsoft.Fx.Portability/ApiPortService.cs
@@ -27,63 +27,12 @@ namespace Microsoft.Fx.Portability
         }
 
         private static readonly TimeSpan Timeout = TimeSpan.FromMinutes(10);
+
         private readonly CompressedHttpClient _client;
 
-        public ApiPortService(string endpoint, ProductInformation info, IProxyProvider proxyProvider)
+        public ApiPortService(string endpoint, ProductInformation info, IProxyProvider proxyProvider = null)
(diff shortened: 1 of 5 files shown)


src/Microsoft.Fx.Portability/ApiPortService.cs
```diff
@@ -219,5 +168,35 @@ public void Dispose()
...
+
+            // Create the URI directly from a string (rather than using a hard-coded scheme or port) because 
+            // even though production use of ApiPort should always use HTTPS, developers using a non-production
+            // portability service URL (via the -e command line parameter) may need to specify a different 
+            // scheme or port.
+            var uri = new Uri(endpoint);
+
+            var clientHandler = new HttpClientHandler
+            {
+#if !FEATURE_SERVICE_POINT_MANAGER
```
conniey: I just noticed this. Do you remember why FEATURE_SERVICE_POINT_MANAGER was only defined for .NET 4.6 but not .NET Standard?

twsouthwick:
I started doing some tests, but the proxy is currently not set up well for testing. I'll get this merged in and then do another PR for better tests around it.

src/Microsoft.Fx.Portability/ApiPortService.cs
twsouthwick: The way .NET Framework and .NET Standard handle SSL is different. @mjrousos could comment better on that

PR #532, opened 2017-12-01
Don't print current culture in VSIX output

### LINKED ISSUE #529: Localized strings in VSIX show locale information
**Repro Steps**
- Set computer locale to en-CA
- Fetch dev or master branch from dotnet-apiport
- Debug ApiPort.Vsix
- Right-click on project -> Analyze Project Portability

**Actual**
image

**Expected**
Expect the strings without the language information in front (ie. en-CA is not there anymore)

chlowell:
Fixes #529

### DIFF
--- a/src/ApiPort.VisualStudio/AnalyzeMenu.cs
+++ b/src/ApiPort.VisualStudio/AnalyzeMenu.cs
@@ -172,8 +172,8 @@ namespace ApiPortVS
             var header = new StringBuilder();
 
             header.AppendLine(string.Format(CultureInfo.CurrentCulture, LocalizedStrings.CopyrightFormat, _productInfo.Version));
-            header.AppendLine(string.Concat(CultureInfo.CurrentCulture, LocalizedStrings.AboutTool, " ", DocumentationLinks.About));
-            header.AppendLine(string.Concat(CultureInfo.CurrentCulture, LocalizedStrings.AboutPrivacy, " ", DocumentationLinks.PrivacyPolicy));
+            header.AppendLine($"{LocalizedStrings.AboutTool} {DocumentationLinks.About}");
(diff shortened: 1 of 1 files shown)


src/ApiPort.VisualStudio/AnalyzeMenu.cs
```diff
@@ -172,8 +172,8 @@ private string[] PromptForAssemblyPaths()
             var header = new StringBuilder();
 
             header.AppendLine(string.Format(CultureInfo.CurrentCulture, LocalizedStrings.CopyrightFormat, _productInfo.Version));
-            header.AppendLine(string.Concat(CultureInfo.CurrentCulture, LocalizedStrings.AboutTool, " ", DocumentationLinks.About));
-            header.AppendLine(string.Concat(CultureInfo.CurrentCulture, LocalizedStrings.AboutPrivacy, " ", DocumentationLinks.PrivacyPolicy));
+            header.AppendLine($"{LocalizedStrings.AboutTool} {DocumentationLinks.About}");
```
conniey: One of the issues we had from FxCop was that we weren't being clear on specifying `CultureInfo.CurrentCulture` and using `string.Format` instead. Did we disable that rule, @twsouthwick

conniey: Nvm, It looks like the dotnet-ci builds are down.

conniey:
@dotnet-bot test this please

src/ApiPort.VisualStudio/AnalyzeMenu.cs
twsouthwick: Interpolated strings by default are culture specific, so this should be fine. The rule you're thinking of is when we need invariant strings

PR #536, opened 2017-12-13
Use System.CommandLine to simplify command line parsing

twsouthwick:


NuGet.config
```diff
@@ -1,9 +1,10 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <!-- Don't use any sources set by the system (workaround for microbuild issue) -->
+     <!-- Don't use any sources set by the system (workaround for microbuild issue) -->
```
conniey: nit: unnecessary space

tests/ApiPort.Tests/AnalyzeOptionsTests.cs
```diff
@@ -1,10 +1,8 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
```
conniey: Just to make sure, do we have tests for the other options like: listTargets, listResultFormat, docId, etc? Also, I didn't see any tests covering options like "Description", "Endpoint", etc.

it would be nice to see, but we can create an issue to add those tests.. in addition to the last set of tests from the proxy PR #531.
I think there was another PR

twsouthwick: We don't. I'll add some basic ones

twsouthwick:
@dotnet-bot please test osx

PR #538, opened 2017-12-15
Update HTML report generation

chlowell:
This makes some changes to `Microsoft.Fx.Portability.Reports.Html`:
 - update RazorEngine dependency to the latest (unofficial) prerelease
 - helpers become inline Razor or `HtmlHelper` methods 
 - compatibility results and summary content become partial views
 - report creation test generates all report sections
 - more string literals moved to resources
 - alt text added to buttons and success/error icons

### DIFF
--- a/src/Microsoft.Fx.Portability.Reports.Html/CompatibilitySummaryModel.cs
+++ b/src/Microsoft.Fx.Portability.Reports.Html/CompatibilitySummaryModel.cs
@@ -9,10 +9,10 @@ namespace Microsoft.Fx.Portability.Reports.Html
 {
     public class CompatibilitySummaryModel
     {
-        public IDictionary<BreakingChange, IEnumerable<MemberInfo>> Breaks { get; private set; }
-        public AssemblyInfo Assembly { get; private set; }
-        public ReportingResult ReportingResult { get; private set; }
-        public int LoopCounter { get; private set; }
+        public IDictionary<BreakingChange, IEnumerable<MemberInfo>> Breaks { get; }
+        public AssemblyInfo Assembly { get; }
+        public ReportingResult ReportingResult { get; }
+        public int LoopCounter { get; }
 
         public CompatibilitySummaryModel(IDictionary<BreakingChange, IEnumerable<MemberInfo>> breaks, AssemblyInfo assembly, ReportingResult reportingResult, int loopCounter)
         {
--- a/src/Microsoft.Fx.Portability.Reports.Html/CompatibilitySummaryModel.cs
+++ b/src/Microsoft.Fx.Portability.Reports.Html/CompatibilitySummaryModel.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Fx.Portability.ObjectModel;
+using Microsoft.Fx.Portability.Reporting.ObjectModel;
+using System.Collections.Generic;
+
+namespace Microsoft.Fx.Portability.Reports.Html
+{
+    public class CompatibilitySummaryModel
+    {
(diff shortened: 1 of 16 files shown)


conniey:
@chlowell Looks like it's having troubles invoking the tests for the HTML writer. Are we still compiling for net461? Maybe we're not referencing the test adapter package for the HTML writer? First thoughts.

```
14:36:04 Test run for D:\j\w\windows_nt_re---1ed4f02d\bin\Release\Microsoft.Fx.Portability.Reports.Html.Tests\net46\Microsoft.Fx.Portability.Reports.Html.Tests.dll(.NETFramework,Version=v4.6)
14:36:04 Microsoft (R) Test Execution Command Line Tool Version 15.3.0-preview-20170628-02
14:36:04 Copyright (c) Microsoft Corporation.  All rights reserved.
14:36:04 
14:36:04 The path 'D:\j\w\windows_nt_re---1ed4f02d\bin\Release\Microsoft.Fx.Portability.Reports.Html.Tests\net46' specified in the 'TestAdapterPath' is invalid. Error: The custom test adapter search path provided was not found, provide a valid path and try again.
14:36:04 The test source file "D:\j\w\windows_nt_re---1ed4f02d\bin\Release\Microsoft.Fx.Portability.Reports.Html.Tests\net46\Microsoft.Fx.Portability.Reports.Html.Tests.dll" provided was not found.
```

src/Microsoft.Fx.Portability.Reports.Html/Resources/_PortabilityReport.cshtml
```diff
@@ -257,13 +218,13 @@
                         {
                             <td>@nugetPackage.AssemblyInfo</td>
                         }
-                      
+
```
twsouthwick: Why the new lines?

chlowell: Each of these is a line of spaces being replaced with a single newline.

src/Microsoft.Fx.Portability.Reports.Html/HtmlReportWriter.cs
```diff
@@ -19,21 +20,18 @@ public class HtmlReportWriter : IReportWriter
...
-            _formatInformation = new ResultFormatInformation
-            {
-                DisplayName = "HTML",
-                MimeType = "text/html",
-                FileExtension = ".html"
-            };
         }
 
-        public ResultFormatInformation Format { get { return _formatInformation; } }
+        public ResultFormatInformation Format { get; } = new ResultFormatInformation
```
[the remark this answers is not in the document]
chlowell: I also think we only ever have a single instance. Even so, the backing field may as well be static.

src/Microsoft.Fx.Portability.Reports.Html/CompatibilitySummaryModel.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.Fx.Portability.Reports.Html
+{
+    public class CompatibilitySummaryModel
+    {
+        public IDictionary<BreakingChange, IEnumerable<MemberInfo>> Breaks { get; }
+        public AssemblyInfo Assembly { get; }
+        public ReportingResult ReportingResult { get; }
+        public int LoopCounter { get; }
```
conniey: What does LoopCounter represent? Is it the number of issues encountered?

chlowell: I created the property while refactoring the report into partial views. Turns out it's unneeded. The counter value was used to ensure unique ids for compatibility summaries so the script which collapses/expands them can work. Now that summaries are generated in a partial view, we can just generate GUIDs there instead of passing a unique identifier in via the model.

PR #540, opened 2018-01-02
Ensure offline ApiPort loads report writers

chlowell:
ApiPort.Offline currently doesn't load report writers (and thus doesn't write reports) because `DependencyBuilder.TryLoadReportWriters` is never called.


src/ApiPort/DependencyBuilder.Offline.cs
```diff
@@ -15,6 +15,7 @@ internal static partial class DependencyBuilder
         static partial void RegisterOfflineModule(ContainerBuilder builder)
         {
             builder.RegisterModule(new OfflineDataModule(DefaultOutputFormatInstanceName));
+            TryLoadReportWriters(builder);
         }
 
         private static void TryLoadReportWriters(ContainerBuilder builder)
```
twsouthwick: The pattern of "Try..." usually returns a Boolean. As part of this change, can you remove that prefix since it doesn't return any status?

PR #546, opened 2018-01-12
Fix issue where Autofac cannot resolve IApiPortService

conniey:
* Console tool could not resolve `ApiPortClient` for commands other than `analyze` and `dump` because IApiPortService was not registered.
* This always registers ApiPortService if the command is not dump
* This also updates some READMEs that had broken links and makes it more concise. Hopefully

### DIFF
--- a/docs/VSExtension/README.md
+++ b/docs/VSExtension/README.md
@@ -8,7 +8,7 @@ __Features__
 
 * Analysis of a single project, solution, or binaries
 * Analysis against multiple .NET platforms simultaneously
-* Generation of multiple analysis reports in different formats. Supported 
+* Generation of multiple analysis reports in different formats. Supported
 formats are:
     * JSON
     * HTML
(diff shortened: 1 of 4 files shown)


docs/VSExtension/README.md
```diff
@@ -98,9 +98,9 @@ A list of changes can be found in the [Changelog](Changelog.md).
...
 [PlatformPortability]: ../HowTo/PlatformPortability.md
-[PrivacyPolicy]: ../LicenseTerms/Microsoft%20.NET%20Portability%20Analyzer%20Privacy%20Statement.txt
+[PrivacyPolicy]: ../LicenseTerms/Microsoft%20.NET%20Portability%20Analyzer%20CLI%20Privacy%20Statement.txt
 [ReportToolWindow]: img/report.toolWindow.png
 [SampleReport]: img/analysisReport.png
 [SolutionExplorer-ContextMenu]: img/analysisContextMenu.png
 [SourceCodeMapping]: img/sourceMapping.output.png
-[VSGallery]: https://visualstudiogallery.msdn.microsoft.com/1177943e-cfb7-4822-a8a6-e56c7905292b
\ No newline at end of file
+[VSGallery]: https://marketplace.visualstudio.com/items?itemName=ConnieYau.NETPortabilityAnalyzer
```
twsouthwick: Is this released under your name?

conniey: Yes. It's because it had to belong to a personal MSDN account and we don't have a team account. So it was moved to me.

PR #547, opened 2018-01-12
Fix Code Analysis warnings

conniey:
This fixes warnings that an upgrade to Microsoft.CodeAnalysis.FxCopAnalyzers 2.6.0 will discover after we migrate to it.

### DIFF
--- a/src/ApiPort/ConsoleProgressReporter.cs
+++ b/src/ApiPort/ConsoleProgressReporter.cs
@@ -11,7 +11,7 @@ using System.Threading.Tasks;
 
 namespace ApiPort
 {
-    public class ConsoleProgressReporter : IProgressReporter
+    public sealed class ConsoleProgressReporter : IProgressReporter
     {
         private readonly List<string> _issuesReported = new List<string>();
         private readonly List<ConsoleProgressTask> _progressTasks = new List<ConsoleProgressTask>();
--- a/src/Microsoft.Fx.Portability/TextWriterProgressReporter.cs
+++ b/src/Microsoft.Fx.Portability/TextWriterProgressReporter.cs
@@ -7,7 +7,7 @@ using System.IO;
 
 namespace Microsoft.Fx.Portability
 {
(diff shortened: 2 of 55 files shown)


src/ApiPort/ConsoleProgressReporter.cs
```diff
@@ -89,17 +89,28 @@ private static void WriteColorLine(string message, ConsoleColor color)
 
         public void Dispose()
         {
-            if (!_disposed)
+            Dispose(true);
```
twsouthwick: Another option here would be to seal the type. Is that possible?

src/Microsoft.Fx.Portability/TextWriterProgressReporter.cs
```diff
@@ -41,6 +42,12 @@ public void Resume()
         { }
 
         public void Dispose()
```
twsouthwick: As mentioned above, maybe sealing the type would work

tests/ApiPort.Tests/AnalyzeOptionsTests.cs
```diff
@@ -31,15 +31,15 @@ public void TestAssemblyFlag_Directory()
...
+        public static void NoArgs()
         {
             var options = CommandLineOptions.ParseCommandLineOptions(Array.Empty<string>());
 
             Assert.Equal(AppCommand.Exit, options.Command);
         }
 
         [Fact]
-        public void AnalyzeNoFile()
+        public static void AnalyzeNoFile()
```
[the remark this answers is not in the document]
conniey: Yes. I thought it was magical too that the tests were still found and running.

src/ApiPort/ConsoleProgressReporter.cs
conniey: Yes. Sealing the type would be a lot easier. I'll do that for both `ProgressReporters`

conniey:
@twsouthwick  I verified that the VS Extension still works in VS2017 with the changes. Is everything OK?

PR #551, opened 2018-01-16
Adds an Advanced Options page

conniey:
* Enables the ability to update and view the service endpoint

image

**Note:** Before we merge, I need to test this out on VS2015


twsouthwick:
I'm going to close this per some offline discussion we've had about not exposing this via the options.

PR #554, opened 2018-01-25
Minor fixes to Excel Report

conniey:
* Changed the Assembly header from "Header for assembly name entries" to "Assembly Name"
* Always include .NET platform targets in Excel summary output


conniey:
@dotnet-bot test Windows_NT Debug please

Issue #494, opened 2017-10-04
build.sh returns no output on Linux


trinaldi:
Following the `README.md`, I cloned the repo and ran `./build.sh`, but nothing - not even an error message - shows up.

I'm running Ubuntu 17.04.
```
uname -a
Linux tiago-linux 4.13.0-994-generic #201709160046 SMP Sat Sep 16 04:48:27 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

```
dotnet --version
2.0.0
```

If i'm missing something, please let me know.

conniey:
@trinaldi Hi. I apologise for the delay. Our tool has only been tested on 16.04 and 14.04. I was trying out your scenario and .NET Core 1.0 SDK is not supported on 17.04, so you would be unable to run the tests as well.

A workaround would be to modify `src/ApiPort/ApiPort.csproj` to support `netcoreapp2.0` rather than netcoreapp1.0, then do a dotnet restore and dotnet build on it.

trinaldi:
Thanks, @conniey .

Maybe it's time to update the tests or create a .NET Core 2.0 SDK branch!

Again, thanks for your reply. I'll try this workaround.

conniey:
@trinaldi I submitted a PR yesterday, it should be fixed in dev and master now. Let me know if it's unsuccessful for you. I tried it in Ubuntu 17.04 and 17.10.

PR #559, opened 2018-01-26
Support .NET Core 2.0

conniey:
Updates our tests to .NET Core App 2.0. This is because this target is available on more distributions.


conniey:
Should fix #494

PR #560, opened 2018-01-26
Merge latest fixes from dev into master

conniey:
* Fixes Code Analysis Warnings
* Updates to use Microsoft.CodeAnalysis.FxCopAnalyzers 2.6.0
* Fixes to Excel headers and adds version numbers to the targets
* Fixes to Breaking Changes analysis
* Updates solution to .NET Core 2.0

### DIFF
--- a/tests/Microsoft.Fx.Portability.Tests/BreakingChangeParserTests.cs
+++ b/tests/Microsoft.Fx.Portability.Tests/BreakingChangeParserTests.cs
@@ -193,15 +193,6 @@ namespace Microsoft.Fx.Portability.Tests
             ValidateParse(GetBreakingChangeMarkdown("MultipleBugLinks.md"), expected);
         }
 
-        [Fact]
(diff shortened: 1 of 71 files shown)


tests/Microsoft.Fx.Portability.Tests/BreakingChangeParserTests.cs
```diff
@@ -191,6 +193,15 @@ public void BreakingChangeMultipleLinks()
             ValidateParse(GetBreakingChangeMarkdown("MultipleBugLinks.md"), expected);
         }
 
+        [Fact]
+        public void CommentsInCategories()
```
chlowell: :suspect: my bad, this test is redundant. Can you please remove it?

conniey: Done!

PR #567, opened 2018-01-31
Fixing FxCop issues

conniey:
* Fixing CA3053 issues with XmlSecureResolver
* Fixing globalisation issues with interpolated strings

### DIFF
--- a/src/Microsoft.Fx.Portability/TargetMapper.cs
+++ b/src/Microsoft.Fx.Portability/TargetMapper.cs
@@ -84,6 +84,8 @@ namespace Microsoft.Fx.Portability
             Load(stream, null);
         }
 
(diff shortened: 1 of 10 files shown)


src/Microsoft.Fx.Portability/TargetMapper.cs
```diff
@@ -114,7 +115,9 @@ private void Load(Stream stream, string path)
                         XmlResolver = null
                     };
 
-                    schemas.Add(null, XmlReader.Create(xsdStream, xmlReaderSettings));
+                    var reader = XmlReader.Create(xsdStream, xmlReaderSettings);
```
twsouthwick: this should be disposed, but not a blocker for now

PR #575, opened 2018-02-22
Add an alias to the docs for convenience

glennblock:
Adds quick tip to the README to use an alias command for convenience


glennblock:
👍

PR #578, opened 2018-03-01
Moved projects into subfolders to prepare for Portability Service work

mjrousos:
ApiPort* projects (and tests) moved into ApiPort folder
Microsoft.Fx* projects (and tests) moved into lib folder

We can use a top-level src/backend folder for the Portability Service components


twsouthwick:
LGTM The only downside of moves like this is it messes with history and makes it hard to follow previous changes :(

twsouthwick:
@mjrousos should this be done in dev and then merged into service from there?

conniey:
Yeah.. thinking more about it, with all the renames, probably a good idea to put the base of this against the dev, branch. create another PR to merge into master and then pull that change into service.

mjrousos:
Yea, that makes sense. Otherwise merging future dev changes into service will be a pain. I'll abandon this and re-create it in dev.

PR #579, opened 2018-03-01
Add netstandard as a possible home for System.Object

mjrousos:
This is needed in case of .NET Standard NetFx facade assemblies
which won't have references to mscorlib or System.Runtime.

Prior to this change, ApiPort was failing when analyzing .NET Framework facade assemblies since they didn't have references to the usual 'base' assemblies (system.runtime, mscorlib, etc.).


twsouthwick:
Can you add a test case similar to the multiple mscorlib example?

mjrousos:
Will do. Also, it occurs to me that I probably ought to make the PR against dev rather than against service since we'll want this to be fixed in parallel with our portability service work.

mjrousos:
I've created a [new branch](https://github.com/mjrousos/dotnet-apiport/tree/NetstandardFixFindObject) with this fix (and a test) relative to Dev but will wait for my other PR to merge first to save merge conflict trouble.

PR #581, opened 2018-03-02
Update SystemObjectFinder to support NetStandard facades

mjrousos:
This is needed in case of .NET Standard NetFx facade assemblies
which won't have references to mscorlib or System.Runtime.


tests/lib/Microsoft.Fx.Portability.MetadataReader.Tests/Tests/OnlyNetStandardReference.il
```diff
@@ -0,0 +1,29 @@
+﻿///
```
twsouthwick: Cool - This is simpler to repro since you don't need to actually have the reference assemblies. I'll take your lead and switch the 2 mscorlib example to IL so it can be compiled without setting a Win7 VM up :)

twsouthwick:
@dotnet-bot please test OSx10.12

twsouthwick:
@dotnet-bot test OSX10.12 Debug please

twsouthwick:
@dotnet-bot test OSX10.12 Release please

twsouthwick:
@mjrousos The OSX failures can be ignored. There's some hardware failure going on with no ETA. We don't need to wait for them to merge this in

PR #583, opened 2018-03-05
Compile IL at runtime

twsouthwick:
This removes any persisted binaries for tests and ensures the IL in the repo is what is being tested


tests/lib/Microsoft.Fx.Portability.MetadataReader.Tests/ManagedMetadataReaderTests.cs
```diff
@@ -226,12 +226,12 @@ public static void MultidimensionalPrimitiveArray()
         }
 
         [Fact]
-        public static void ThrowsSystemObjectNotFoundException()
+        public void ThrowsSystemObjectNotFoundException()
         {
             var dependencyFilter = Substitute.For<IDependencyFilter>();
             dependencyFilter.IsFrameworkAssembly(Arg.Any<AssemblyReferenceInformation>()).Returns(false);
             var dependencyFinder = new ReflectionMetadataDependencyFinder(dependencyFilter, new SystemObjectFinder(dependencyFilter));
-            var assemblyToTest = TestAssembly.Create("FilterApis.cs", false, new[] { typeof(Image).GetTypeInfo().Assembly.Location });
```
mjrousos: The additional reference isn't needed anymore?

tests/lib/Microsoft.Fx.Portability.MetadataReader.Tests/TestAssembly.cs
```diff
@@ -124,5 +121,74 @@ public ResourceStreamAssemblyFile(string fileName)
...
+                    _output.WriteLine(stdout);
+
+                    _output.WriteLine("ilasm stderr:");
+                    _output.WriteLine(stderr);
+
+                    Assert.Equal(0, process.ExitCode);
+                }
+
+                File.Delete(tmp);
+                var output = Path.ChangeExtension(tmp, ".dll");
```
mjrousos: The name `output` is a little confusing with `_output`; maybe rename this to outputAssembly or something like that so they're not too similar?

tests/lib/Microsoft.Fx.Portability.MetadataReader.Tests/TestAssembly.cs
```diff
@@ -5,38 +5,37 @@
...
         {
-        switch (Path.GetExtension(source).ToLowerInvariant())
+            switch (Path.GetExtension(source).ToLowerInvariant())
             {
                 case ".dll":
                 case ".exe":
                     return new ResourceStreamAssemblyFile(source);
                 case ".cs":
-                    return new CSharpCompileAssemblyFile(source, allowUnsafe, additionalReferences);
+                    return new CSharpCompileAssemblyFile(source, allowUnsafe, Enumerable.Empty<string>());
```
mjrousos: Not urgent, but it might be nice to write Roslyn diagnostics to output in this case since we'll always have an ITestOutputHelper anyhow.

mjrousos: Also, do we not have a way to provide additional assembly reference for C# tests anymore?

tests/lib/Microsoft.Fx.Portability.MetadataReader.Tests/TestAssembly.cs
```diff
@@ -124,5 +121,74 @@ public ResourceStreamAssemblyFile(string fileName)
...
+                };
+
+                using (var process = Process.Start(psi))
+                {
+                    process.WaitForExit();
+
+                    var stdout = process.StandardOutput.ReadToEnd();
+                    var stderr = process.StandardError.ReadToEnd();
+
+                    _output.WriteLine("ilasm stdout:");
```
mjrousos: It might be nice to make `_output` optional and only write to it if it's non-null. It's not actually required.

tests/lib/Microsoft.Fx.Portability.MetadataReader.Tests/xunit.runner.json
```diff
@@ -1,4 +1,5 @@
 ﻿{
   "methodDisplay": "method",
-  "diagnosticMessages": true
-}
\ No newline at end of file
+  "diagnosticMessages": true,
+  "shadowCopy": false
```
mjrousos: Is necessary so that ilasm will be present?

tests/lib/Microsoft.Fx.Portability.MetadataReader.Tests/ManagedMetadataReaderTests.cs
twsouthwick: Nope. Not sure why it was there. We actually don't need it for anything.

tests/lib/Microsoft.Fx.Portability.MetadataReader.Tests/TestAssembly.cs
twsouthwick: I can add that back, we just currently don't need it so I removed it

tests/lib/Microsoft.Fx.Portability.MetadataReader.Tests/xunit.runner.json
twsouthwick: Yup. Since it's never actually referenced by the tests, so it doesn't get copied by xunit :(  I tried adding a comment in there, but it fails to parse with comments

tests/lib/Microsoft.Fx.Portability.MetadataReader.Tests/TestAssembly.cs
twsouthwick: It's not, but it'll make it easier to see if there's an error and I'd rather default to having the logs if needed by forcing it to be there.

tests/lib/Microsoft.Fx.Portability.MetadataReader.Tests/TestAssembly.cs
twsouthwick: Added the diagnostics to the output window

PR #585, opened 2018-03-06
Remove unnecessary update file

twsouthwick:
After the recent change to compile the IL at runtime, this is now unnecessary. Forgot to remove it with the previous PR


mjrousos:
LGTM

PR #589, opened 2018-03-07
Add Analyze function and tests

chlowell:
Adds a project for Azure Functions with one function. That function returns guids for ApiPort analysis requests (i.e. posts with gzipped json for a single `AnalyzeRequest` as content). Anything else gets 400.


src/backend/Functions/Analyze.cs
```diff
@@ -0,0 +1,65 @@
...
+            response.Content = new StringContent(submissionId);
+
+            return response;
+        }
+
+        public static async Task<AnalyzeRequest> DeserializeRequest(HttpContent content)
+        {
+            try
+            {
+                var body = await DecompressContent(content);
```
conniey: It's better to deserialize from stream rather than a string because we used to have a bunch of exceptions for strings being too long when the analysis request was huge.
I believe there is an extension [`DataExtensions.Deserialize<T>`](https://github.com/Microsoft/dotnet-apiport/blob/dev/src/lib/Microsoft.Fx.Portability/DataExtensions.cs#L21) that may do what you want.

PR #590, opened 2018-03-09
Add API Gateway service

mjrousos:
This simple microservice acts as a single point of entry for clients like ApiPort. Its primary purpose is to reroute calls to different downstream services base on request path.

It also adds some reliability by caching common results, rate limiting more expensive POST operations, and adding QoS policies like a circuit breaker (though the circuit breaker isn't quite as robust as I would like due to Ocelot limitations; this story should be improved soon at which point we can update the gateway).


mjrousos:
Tests are coming soon, but I figured I'd make this PR now to start conversation since I may be sidetracked for a few days with other work.

src/backend/PortabilityService.Gateway/appsettings.json
```diff
@@ -0,0 +1,269 @@
+﻿{
+  "ApplicationInsights": {
+    "InstrumentationKey": "<AI-Key-Goes-Here>"
+  },
+  "Gateway": {
```
mjrousos: I'm debating moving the Gateway/ReRoutes config to its own file since it is so large. I'm not sure if that would be helpful or not.

src/backend/PortabilityService.Gateway/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            // Create Serilog configuration from app configuration
+            var serilogLoggerConfiguration = new LoggerConfiguration()
+                .ReadFrom.Configuration(config);
+            
+            if (!env.IsDevelopment())
+            {
+                // In non-dev environments, add an App Insights sink
+                serilogLoggerConfiguration = serilogLoggerConfiguration
+                                                .WriteTo.ApplicationInsightsTraces(config[APP_INSIGHTS_KEY_KEY]);
```
twsouthwick: nit indentation

src/backend/PortabilityService.Gateway/Reliability/PortabilityServiceQoSProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+    /// that and this type should be updated according to the comments below once that's done.
+    /// </summary>
+    public class PortabilityServiceQoSProvider : IQoSProvider
+    {
+        private readonly ILogger _logger;
+        private readonly CircuitBreaker _circuitBreaker;
+
+        public PortabilityServiceQoSProvider(DownstreamReRoute reRoute, ILoggerFactory loggerFactory)
+        {
+            _logger = loggerFactory.CreateLogger<PortabilityServiceQoSProvider>();
```
twsouthwick: Can you just ask for a `ILogger<PortabilityServiceQoSProvider>` in the constructor arg?

src/backend/PortabilityService.Gateway/Reliability/PortabilityServiceQoSProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+                .CircuitBreakerAsync(options.ExceptionsAllowedBeforeBreaking, TimeSpan.FromMilliseconds(options.DurationOfBreak),
+                (exception, duration) =>
+                {
+                    // Log warning when circuit break is opened
+                    _logger.LogWarning("Circuit breaker opened for {circuitBreakerDuration}ms due to: {error}",
+                        duration.TotalMilliseconds,
+                        $"[{exception.GetType().Name}] {exception.Message}");
+                    if (exception != null)
+                    {
+                        _logger.LogTrace("Full exception: {exception}", exception.ToString());
```
twsouthwick: This seems like it should be included with the previous warning as it would allow correlating with the structure logging

src/backend/PortabilityService.Gateway/Reliability/PortabilityServiceQoSProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+                //        if we handle a non-exceptional response type and one TimeoutPolicy. I should look into updating it 
+                //        to use an arbitrary PolicyWrap or at least IEnumerable<IPolicy>.
+                // .OrResult<HttpResponseMessage>(response => ReportsServerError(response))
+                .CircuitBreakerAsync(options.ExceptionsAllowedBeforeBreaking, TimeSpan.FromMilliseconds(options.DurationOfBreak),
+                (exception, duration) =>
+                {
+                    // Log warning when circuit break is opened
+                    _logger.LogWarning("Circuit breaker opened for {circuitBreakerDuration}ms due to: {error}",
+                        duration.TotalMilliseconds,
+                        $"[{exception.GetType().Name}] {exception.Message}");
```
twsouthwick: Logging is done with structured logging, so you shouldn't need to access specific properties unless needed. Here, passing in `duration` and `exception` would be better

src/backend/PortabilityService.Gateway/Reliability/PortabilityServiceQoSProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+                {
+                    // Log informational message when the circuit breaker is half-opem
+                    _logger.LogInformation("Circuit half-open");
+                });
+
+            // Stop waiting on requests after a period of time
+            var timeoutPolicy = Policy.TimeoutAsync(TimeSpan.FromMilliseconds(options.TimeoutValue), options.TimeoutStrategy,
+                (context, timespan, timedOutAction) =>
+                {
+                    _logger.LogWarning("Downstream call timed out after {timeout}ms", timespan.TotalMilliseconds);
```
twsouthwick: same as above with structured logging, you can just pass `timespan`

src/backend/PortabilityService.Gateway/Startup.cs
```diff
@@ -0,0 +1,87 @@
...
+using PortabilityService.Gateway.Reliability;
+using System.Linq;
+
+namespace PortabilityService.Gateway
+{
+    public class Startup
+    {
+        const string HEALTH_CHECK_PATH = "/hc";
+
+        public Startup(IConfiguration configuration, ILoggerFactory loggerFactory)
```
twsouthwick: nit: unnecessary arg

src/backend/PortabilityService.Gateway/appsettings.json
twsouthwick: I like the idea of a separate `appsettings.gateway.json` due to the size

src/backend/PortabilityService.Gateway/Startup.cs
```diff
@@ -0,0 +1,87 @@
...
+            });
+
+            var ocelotConfiguration = new OcelotPipelineConfiguration
+            {
+                // Work around https://github.com/ThreeMammals/Ocelot/issues/263
+                PreQueryStringBuilderMiddleware = ContentHeaderForwardingMiddleware.ForwardContentHeaders
+            };
+
+            // Ocelot must be the last middleware in the pipeline
+            app.UseOcelot(ocelotConfiguration).Wait();
```
twsouthwick: I think you can make the Configure async by returning a Task from it instead of using `.Wait()`

src/backend/PortabilityService.Gateway/appsettings.json
```diff
@@ -0,0 +1,269 @@
...
+    "GlobalConfiguration": {
+      "RequestIdKey": "Correlation-Id"
+    }
+  },
+  "Serilog": {
+    "MinimumLevel": "Warning",
+    "WriteTo": [
+      {
+        "Name": "Console",
+        "Args": {
```
jeremymeng: We probably don't want to log to Console in production.

src/backend/PortabilityService.Gateway/PortabilityService.Gateway.csproj
```diff
@@ -0,0 +1,27 @@
...
+        (as opposed to a library) to need a strong name signature -->
+    <SignAssembly>false</SignAssembly>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.2.1" />
+    <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.5" />
+    <PackageReference Include="ocelot" Version="3.1.9" />
+    <PackageReference Include="Polly" Version="5.3.1" />
+    <PackageReference Include="Serilog.AspNetCore" Version="2.1.0" />
```
jeremymeng: Is Polly used directly? If it's pulled in by ocelot we could remove it.

src/backend/PortabilityService.Gateway/Reliability/PortabilityServiceQoSProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+            _circuitBreaker = CreateOcelotCircuitBreaker(reRoute.QosOptionsOptions);
+        }
+
+        private CircuitBreaker CreateOcelotCircuitBreaker(QoSOptions options)
+        {
+            // Retry failed requests
+            // TODO : Ocelot does not currently allow more policies to be added.
+            //        Update Ocelot to allow this and then add this policy.
+            var _retryCount = 3; // options.RetryCount;
+            var _retryBaseDelay = 0.5; // options.RetryBaseDelay
```
jeremymeng: nit: remove underscores for locals

src/backend/PortabilityService.Gateway/Reliability/PortabilityServiceQoSProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+                () =>
+                {
+                    // Log informational message when the circuit breaker resets
+                    _logger.LogInformation("Circuit breaker closed");
+                },
+                () =>
+                {
+                    // Log informational message when the circuit breaker is half-opem
+                    _logger.LogInformation("Circuit half-open");
+                });
```
jeremymeng: nit: indent the argument to `.CircuitBreakerAsync()`

src/backend/PortabilityService.Gateway/Reliability/PortabilityServiceQoSProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+                    }
+                },
+                () =>
+                {
+                    // Log informational message when the circuit breaker resets
+                    _logger.LogInformation("Circuit breaker closed");
+                },
+                () =>
+                {
+                    // Log informational message when the circuit breaker is half-opem
```
jeremymeng: >opem [](start = 82, length = 4)

typo

src/backend/PortabilityService.Gateway/appsettings.json
mjrousos: Sounds good. I'll move it.

src/backend/PortabilityService.Gateway/Program.cs
mjrousos: Thanks; fixed.

src/backend/PortabilityService.Gateway/Reliability/PortabilityServiceQoSProvider.cs
mjrousos: Unfortunately we can't because `PortabilityServiceQoSProviderFactory` is what's registered with DI and it instantiates instances of this type. So, we get the `ILoggerFactory` when the factory is created and then pass it along to the type to make its own logger.

src/backend/PortabilityService.Gateway/Reliability/PortabilityServiceQoSProvider.cs
mjrousos: Yea, I had it that way originally. Including the full exception stack makes the warning really long, though, so I split up the exception type and message (which are in the warning) from the full stack (which is logged separately as a trace).

You're probably right, though, that it would be good to have them together. I can go back to including the whole exception string in the warning.

src/backend/PortabilityService.Gateway/Reliability/PortabilityServiceQoSProvider.cs
mjrousos: Good point. Fixed.

src/backend/PortabilityService.Gateway/Reliability/PortabilityServiceQoSProvider.cs
mjrousos: Fixed.

src/backend/PortabilityService.Gateway/Startup.cs
mjrousos: Good catch. Fixed.

src/backend/PortabilityService.Gateway/Startup.cs
mjrousos: You can't, unfortunately. Configure has to return void (aspnet/hosting#29). We could make it async void, but that seems like a bad idea.

src/backend/PortabilityService.Gateway/PortabilityService.Gateway.csproj
mjrousos: It is because I'm making my own policies to use. By default, we can use Ocelot-produced Polly policies (and then wouldn't need the Polly dependency), but I'm replacing the default policies with our own.

src/backend/PortabilityService.Gateway/appsettings.json
mjrousos: I think as long as the logging is minimal (warning, in this case), console logging in production is ok. A lot of orchestrators will collect standard out from their services and make it available.

The console logger used to be quite slow (so this would be bad), but that was fixed in 2.0. And this is Serilog console logging, anyhow, which doesn't have the same perf problem.

src/backend/PortabilityService.Gateway/Reliability/PortabilityServiceQoSProvider.cs
mjrousos: Good catch. FIxed.

src/backend/PortabilityService.Gateway/Reliability/PortabilityServiceQoSProvider.cs
mjrousos: Good catch. Fixed.

src/backend/PortabilityService.Gateway/Reliability/PortabilityServiceQoSProvider.cs
mjrousos: Fixed.

mjrousos:
I've added tests for the API gateway

src/backend/PortabilityService.Gateway/Startup.cs
twsouthwick: Yeah, we don't want to remove async void :) Looks like we're stuck with Wait. We may want to reopen that issue since now there's a use case and it's considered bad practice to directly call `.Wait()` on a task.

PR #592, opened 2018-03-13
User Stories 566983\571103: Workflow queue messaging

nicolehaugen79:
This PR completes the work described in [user story 566983](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/566983) and part of the work for [user story 571103](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/566983).

Specific changes include:
1. Added Workflow Manager that manages the stages for the workflow queue (stored in a storage queue) and invokes calls to underlying services for each stage in the workflow (such as the analysis engine).  As part of this design, the services are loosely coupled to allow easy changes to APIs as needed.
2. Updated the Analyze function to add messages to the workflow queue.
3. Added a ProcessWorkflowQueue function that processes messages in the queue.
4. Added unit tests.

One other subtle change that I made is to have the functions pass in an ILogger instance instead of Tracewriter.  ILogger provides better structural logging which is why it is recommended over TraceWriter.

There are also several tasks that still need to be completed - these will be addressed in the next sprint:
1. Handle queue errors (poison message queue)
2. Better logging and error handling in general

Here is the [detailed design for this work](https://microsoft.sharepoint.com/teams/DevDivCustomerEngagement/_layouts/15/WopiFrame.aspx?sourcedoc={3b1775ea-bbf4-4c8a-9299-2eb368bc7390}&action=edit&wd=target%28%2F%2FModernization.one%7Cc9150b50-339d-463c-852c-2f89c2c7619f%2FQueue%5C%5CWorkflow%20Design%7C54d07347-b34f-462a-9ff9-f6fcbe7b86e3%2F%29&wdorigin=onocopylinktopage)


twsouthwick:
This should be merged into the service branch instead of the dev branch

twsouthwick:
GitHub let me retarget against the service branch :)

WorkflowManagement/WorkflowManagement.csproj
```diff
@@ -0,0 +1,60 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
twsouthwick: Can you use the new csproj file type for this? That way we can retarget to .net core/.net framework much easier.

WorkflowManagement/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
twsouthwick: With the new csproj format, this is not needed and can be removed

WorkflowManagement/WorkflowManager.cs
```diff
@@ -0,0 +1,74 @@
...
+        Telemetry,
+        Finished
+    }
+
+    /// <summary>
+    /// Workflow stages are in the following order: Analyze -> Report -> Telemetry -> Finished
+    /// </summary>
+    public class WorkflowManager
+    {
+        IWorkflowAction[] actions;
```
twsouthwick: Fields should be readonly unless you specifically need to mutate them

WorkflowManagement/WorkflowManager.cs
```diff
@@ -0,0 +1,74 @@
...
+        static WorkflowManager manager;
+
+        public static WorkflowManager GetInstance(IWorkflowActionFactory actionFactory)
```
twsouthwick: We us DI throughout the system so methods to manage lifetime are unnecessary

WorkflowManagement/WorkflowQueueMessage.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace WorkflowManagement
+{
+    public class WorkflowQueueMessage
+    {
+        public WorkflowStage Stage { get; set; }
+        public string SubmissionId { get; set; }
+
```
twsouthwick: nit new line

tests/backend/Functions.Tests/Mock/MockCollector.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
```
twsouthwick: I believe we are already using NSubstitute for mocking within the project. Instead of creating manual mocks like this, you can just use that. ie `var collector = Substitute.For<ICollector<T>>()`

PortabilityTools.sln
```diff
@@ -92,7 +92,9 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "backend", "backend", "{0189
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "backend", "backend", "{DE3F5141-1F37-45F6-9A63-697E05E9BF8F}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Functions.Tests", "tests\backend\Functions.Tests\Functions.Tests.csproj", "{CE644844-BE5D-4B95-9C3E-4DEDFCD19C48}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Functions.Tests", "tests\backend\Functions.Tests\Functions.Tests.csproj", "{CE644844-BE5D-4B95-9C3E-4DEDFCD19C48}"
+EndProject
```
jeremymeng: For reference I found this issue. It seems that 9A19103F-16F7-4668-BE54-9A1E7A4F7556 is the one from latest VS while FAE04EC0-... is a legacy one.  Maybe we should all upgrade our VS to the latest in case old versions revert back to FAE04EC0...

src/backend/Functions/Analyze.cs
```diff
@@ -1,38 +1,50 @@
...
 {
     public static class Analyze
     {
+        //Allows ActionFactory to be "injected" as needed, such as with a mock action factory when the function is called through tests
+        public static Func<IWorkflowActionFactory> GetActionFactory { get; set; } = () => new WorkflowActionFactory();
+
         [FunctionName("analyze")]
         public static async Task<HttpResponseMessage> Run(
             [HttpTrigger(AuthorizationLevel.Function, "post", Route = null)] HttpRequestMessage req,
-            TraceWriter log)
```
jeremymeng: Do we need the trace writer to see azure function output in the azure portal?

nicolehaugen79: No, the recommendation that I saw is that instead for logging, we should be using ILogger since it has better, structural logging: https://docs.microsoft.com/en-us/azure/azure-functions/functions-monitoring#write-logs-in-c-functions

PortabilityTools.sln
jeremymeng: [https://github.com/dotnet/project-system/issues/2832](https://github.com/dotnet/project-system/issues/2832)

---
In reply to: [174208896](https://github.com/Microsoft/dotnet-apiport/pull/592#discussion_r174208896) [](ancestors = 174208896)

jeremymeng:
UpgradeLog.htm should be removed

jeremymeng:
Functions.zip should not be added?

src/backend/Functions/Analyze.cs
jeremymeng: I was referring to the Logs pane in the functions dashboard. Maybe ILogger would just work. Anyway not a big deal.
image

WorkflowManagement/WorkflowManager.cs
nicolehaugen79: We talked about this over IM - recapping for everyone's awareness: For now, will leave as a service locator pattern (which is an anti-pattern).  Connie will add a user story to the backlog to add proper dependency management for functions.

src/backend/Functions/Analyze.cs
```diff
@@ -1,38 +1,50 @@
...
+            [Queue("apiportworkflowqueue")]ICollector<WorkflowQueueMessage> workflowMessageQueue, ILogger log)
```
twsouthwick: Can you move the ILogger to its own line for consistency?

WorkflowManagement/WorkflowManager.cs
nicolehaugen79: I simplified my code a bit so that this singleton is no longer needed...so DI isn't needed for the time being.

src/backend/Functions/Analyze.cs
```diff
@@ -1,38 +1,51 @@
...
             if (analyzeRequest == null)
             {
-                log.Error("invalid request");
+                log.LogError("invalid request");
                 return req.CreateResponse(HttpStatusCode.BadRequest);
             }
 
             var submissionId = Guid.NewGuid().ToString();
-            log.Info($"created submission id {submissionId}");
+            log.LogInformation($"created submission id {submissionId}");
```
twsouthwick: If this is structured logging, then you need to provide a an argument at the end instead of using interpolated strings. ie `log.LogInformation("Created submission id {SubmissionId}", submissionId);`. That way, things such as the submissionId can be correlated later. See https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?tabs=aspnetcore2x#log-message-template for more info

src/backend/Functions/ProcessWorkflowQueue.cs
```diff
@@ -0,0 +1,35 @@
...
+using Microsoft.Azure.WebJobs;
+using Microsoft.Extensions.Logging;
+using WorkflowManagement;
+
+namespace Functions
+{
+    public static class ProcessWorkflowQueue
+    {
+        //Allows WorkflowManager to be "injected" as needed, such as with a mock action factory when the function is called through tests
+        public static Func<WorkflowManager> GetWorkflowManager { get; set; } = () => new WorkflowManager();
```
twsouthwick: Is there a reason to have 2 static instances? Also, do they need to be funcs or is WorkflowManager designed to be a static? It would be better to at least follow the service locator plan until we can get DI set up here. That is, created a static class (ie `ServiceLocator`) that has a single static property `WorkFlowManager` that returns it (a single instance or a shared instance depending on what needs to be done).

src/backend/WorkflowManagement/WorkflowQueueMessage.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace WorkflowManagement
+{
+    public class WorkflowQueueMessage
```
twsouthwick: This is fine for now, but it would be nice to set up the workflow so that we use immutable objects (we could then switch to something like `readonly struct`).

tests/backend/Functions.Tests/AnalyzeTests.cs
```diff
@@ -37,10 +41,14 @@ public class AnalyzeTests
...
             request.Content.Headers.Add("Content-Type", "application/json");
 
-            var response = await Analyze.Run(request, new DoNothingTraceWriter());
+            var workflowQueue = Substitute.For<ICollector<WorkflowQueueMessage>>();
+            var response = await Analyze.Run(request, workflowQueue, NullLogger.Instance);
             var body = await response.Content.ReadAsStringAsync();
 
-            Assert.True(Guid.TryParse(body, out var _));
+            Guid submissionId;
+            Assert.True(Guid.TryParse(body, out submissionId));
```
twsouthwick: you can now do `Assert.True(Guid.TryParse(body, out var submissionId))`. However, since you don't need the result, you can do `Assert.True(Guid.TryParse(body, out _))` and the compiler will even do some optimizations (not important for testing, but potentially in production code in certain circumstances)

src/backend/Functions/Functions.csproj
```diff
@@ -4,10 +4,13 @@
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="2.1.0" />
     <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="1.0.6" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="1.0.6" />
```
chlowell: This reference is in the above ItemGroup as well.

twsouthwick:
@nicolehaugen79 We force squashing in the branch. GitHub does some weird stuff with commit messages by including all the intermediate commits which make the final commit message have unnecessary information. Please verify the message carefully before committing.

tests/backend/Functions.Tests/AnalyzeTests.cs
```diff
@@ -37,10 +41,14 @@ public class AnalyzeTests
...
+
+            workflowQueue.Received().Add(Arg.Is<WorkflowQueueMessage>(x => x.SubmissionId == submissionId.ToString() && x.Stage == WorkflowStage.Analyze));
```
chlowell: This should be a separate test. The function should return the guid in its response, and also queue a message.

src/backend/Functions/Analyze.cs
```diff
@@ -1,38 +1,51 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using Microsoft.Azure.WebJobs;
-using Microsoft.Azure.WebJobs.Extensions.Http;
-using Microsoft.Azure.WebJobs.Host;
-using Microsoft.Fx.Portability;
-using Microsoft.Fx.Portability.ObjectModel;
 using System;
+using System.IO;
+using System.IO.Compression;
```
jeremymeng: Remove unused usings

src/backend/Functions/ProcessWorkflowQueue.cs
```diff
@@ -0,0 +1,35 @@
...
+        public static async Task Run([QueueTrigger("apiportworkflowqueue")]WorkflowQueueMessage workflowMessage, 
+            [Queue("apiportworkflowqueue")]ICollector<WorkflowQueueMessage> workflowMessageQueue, 
+            ILogger log)
+        {
+            log.LogInformation($"processing message {workflowMessage.SubmissionId}, stage {workflowMessage.Stage}");
+
+            var workflowMgr = GetWorkflowManager();
+            var nextMsg = await workflowMgr.ExecuteActionsToNextStage(workflowMessage);
+
+            log.LogInformation($"queueing new message {workflowMessage.SubmissionId}, stage {nextMsg.Stage}");
```
mjrousos: Nit-pick - may want to move this inside the following 'if' since we don't queue anything if `nextMsg.Stage` is `Finished`. You could have an else that just logs that we're done.

src/backend/WorkflowManagement/WorkflowActions/IWorkflowAction.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Tasks;
+
+namespace WorkflowManagement
+{
+    public interface IWorkflowAction
+    {
+        Task Execute(string submissionId);
```
mjrousos: ExecuteAsync instead of Execute, perhaps since it's async?

src/backend/WorkflowManagement/WorkflowManager.cs
```diff
@@ -0,0 +1,75 @@
...
+    public class WorkflowManager
+    {
+        readonly IWorkflowAction[] actions;
+        static WorkflowManager manager;
+
+        public WorkflowManager()
+        {
+            IWorkflowAction[] workflowActions = new IWorkflowAction[3];
+
+            workflowActions[(int)WorkflowStage.Analyze] = new AnalyzeAction();
```
mjrousos: Instead of hard-coding these, it might be better to have `IWorkflowAction` implementations report which stages they handle (through a method or property added to the interface) so that we can add new workflow actions (which will eventually be made available through DI, hopefully) without having to update WorkflowManager logic. Just tell it to invoke the first workflow manager implementation that can handle the stage we're on.

src/backend/WorkflowManagement/WorkflowManager.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// returns a message for the next stage in the queue. 
+        /// </summary>
+        public async Task<WorkflowQueueMessage> ExecuteActionsToNextStage(WorkflowQueueMessage currentMsg)
+        {
+            //Execute the action for the current stage
+            await actions[(int)currentMsg.Stage].Execute(currentMsg.SubmissionId);
+
+            //Determine which stage in the workflow the submission moves to next
+            WorkflowStage nextStage;
+            switch (currentMsg.Stage)
```
mjrousos: Similar to my last comment, I wouldn't include this switch statement. Instead, I think `IWorkflowAction.Execute` should return a `Task<WorkflowStage>` indicating which stage comes next. That way it's the *workflow stage's* responsibility to define the flow, the workflow manager just follows it. It will also allow for more complex scenarios in the future where we don't execute each stage linearly. For example, if the analysis stage failed, it could report that next is the 'ErrorHandling' stage or something instead of moving onto the report stage like usual.

src/backend/Functions/ProcessWorkflowQueue.cs
nicolehaugen79: The reason I made WorkflowManager be static is for testability - so that when used within the context of tests, the WorkflowManager is instantiated using mock workflow actions prior to the Azure Function executing.  Since WorkflowManager is static, when the Azure Function runs, a new instance of WorkflowManager isn't re-created with real actions.  I was using the Service Locator pattern before, but since we didn't have DI, I thought this would be a simpler approach for now.  Anyway, I can switch it back to using Service Locator pattern.

src/backend/WorkflowManagement/WorkflowQueueMessage.cs
nicolehaugen79: Easy enough change to do now - done. :)

tests/backend/Functions.Tests/AnalyzeTests.cs
nicolehaugen79: In this case the guid and queue are linked to the same function\action - so they're not that independent.  Also, in the actual test code, my verification simply adds one line to the test case to verify the queue message.  My suggestion is to leave this as-is for now.  If it proves to be an issue down the road, we can change if needed.  Thoughts?

src/backend/Functions/Analyze.cs
nicolehaugen79: Thought I had removed all these - looks like I missed a few - thanks!

src/backend/Functions/ProcessWorkflowQueue.cs
nicolehaugen79: agree - good catch.

src/backend/WorkflowManagement/WorkflowManager.cs
nicolehaugen79: good suggestion - I have made both this change and the next one.

src/backend/WorkflowManagement/WorkflowManager.cs
```diff
@@ -21,28 +22,53 @@ public class WorkflowManager
...
+            }
+
+            return manager;
+        }
+
+        private WorkflowManager()
+        {
+            actions = new IWorkflowAction[Enum.GetValues(typeof(WorkflowStage)).Length-1];
+
+            AddAction<AnalyzeAction>();
```
mjrousos: Eventually it would be good to get these from DI.

.editorconfig
```diff
@@ -1,8 +0,0 @@
-# 4 space indentation
-[*.cs]
-indent_style = space
-indent_size = 4
-
-[*.ps1]
-indent_style = space
-indent_size = 4
```
mjrousos: Is there a reason for removing this?

nicolehaugen79: I overlooked that I had inadvertently removed this - this is added back now.

Issue #596, opened 2018-03-23
MSBuild 15.1 Issue unknown switch /bl


nmbro:
https://github.com/Microsoft/dotnet-apiport/blob/8759eb488336d12bb1765ba30e0443da11c13073/build.ps1#L131

The binary logging switch is not implemented in MSBuild 15.1 which is a part of dotnet-sdk-2.1.10.101-win-x64

PR #597, opened 2018-03-23
Added ability to detect msbuild version - and set binarylogging if

nmbro:
compatible
Fixes Microsoft/dotnet-apiport#596


conniey:
Thanks a lot for doing this! 👍

PR #599, opened 2018-04-04
Assembly and API Filter Improvements

conniey:
* Only send up and recognise APIs that are part of the .NET platform
* This will help us exclude external libraries like Azure, Application Insights, etc.


tests/lib/Microsoft.Fx.Portability.Tests/Analyzer/DotNetFrameworkFilterTests.cs
```diff
@@ -39,10 +39,10 @@ public void DotNetFrameworkFilterCheckPublicKeyToken(string publicKeyToken, bool
...
         [InlineData("system.something", true)]
-        [InlineData("Microsoft.something", true)]
-        [InlineData("microsoft.something", true)]
-        [InlineData("mono.something", true)]
-        [InlineData("Mono.something", true)]
+        [InlineData("Microsoft.something", false)]
+        [InlineData("microsoft.something", false)]
+        [InlineData("mono.something", false)]
+        [InlineData("Mono.something", false)]
         [InlineData("mscorlib", true)]
```
twsouthwick: Can you add tests for the new prefixes you added?

PR #603, opened 2018-04-06
Added better error handling and poison workflow queue

nicolehaugen79:
I made the following changes:

1.) This is a non-code related change, but communicating it here for awareness.  With AI enabled, the recommendation is to disable the built-in function logging that uses Azure storage since it's not designed for production usage levels: [https://docs.microsoft.com/en-us/azure/azure-functions/functions-monitoring#disable-built-in-logging](https://docs.microsoft.com/en-us/azure/azure-functions/functions-monitoring#disable-built-in-logging).  To turn this off, I deleted the AzureWebJobsDashboard app setting in the portal.

2.) I talked this over with Connie - we decided that when an analysis submission fails, that the end user will be notified of the failure and they are expected to resubmit.  In other words, we currently won't do any automatic retry when a failure occurs.  Based on this decision, I configured the maxDequeueCount to 1 which means that a message will only be dequeued from the incoming storage queue once.

3.) Added a ProcessWorkflowQueuePoision function that logs failures for failed messages and provides a hook point for the future work item to notify the client when a failure occurred.  The ProcessWorkflowQueue function will automatically put a failed message in the poison queue whenever an unhandled exception occurs.  As you'll see from the next bullet, this also includes when the function times out. 

4.) Added CancellationToken handling for the scenario when the function times out.  Based on guidance from the Azure Function team, when the function times out, you need to explicitly handle the cancellation token in your code.  When the token is cancelled, I throw an exception so that the function completes with a failure, the message is dequeued from the incoming storage queue and enqueued into the poison queue.  I also am passing the CancellationToken to the ExecuteAsync methods so that invoked services can do any necessary cleanup (if needed) when a timeout occurs.

5.) I explicitly set the timeout to be 5 minutes.  This is the default, but I wanted to make this apparent in case we need to change it in the future.

6.) Improved the AI trace logging in places.  We can then use Analytics for example, to query traces on the current submission ID to see all of the phases of the workflow that a submission went through.  For example, here's a trace where a failure occurred:

2018-04-04T22:08:45.875 | Created submission id 008dd1c5-2aaf-4509-a192-5d0b2e0c401f | analyze
 
2018-04-04T22:08:46.108 | Queuing new message 008dd1c5-2aaf-4509-a192-5d0b2e0c401f,   stage Analyze | ProcessWorkflowQueue

2018-04-04T22:08:46.313 | Processing message 008dd1c5-2aaf-4509-a192-5d0b2e0c401f, stage   Analyze, id ab8a45dd-1a7a-4ef5-9ba5-e6736821706b | Analyze

2018-04-04T22:08:46.686 | Workflow message failure 008dd1c5-2aaf-4509-a192-5d0b2e0c401f,   stage Analyze | ProcessWorkflowPoisonQueue


conniey:
@twsouthwick , @mjrousos , @chlowell Would one of you two be able to take a look at this? I am afraid this area is out of my expertise other than the usual C#.

src/backend/Functions/ProcessWorkflowQueue.cs
```diff
@@ -14,17 +15,43 @@ public static class ProcessWorkflowQueue
...
         public static async Task Run([QueueTrigger("apiportworkflowqueue")]WorkflowQueueMessage workflowMessage,
             [Queue("apiportworkflowqueue")]ICollector<WorkflowQueueMessage> workflowMessageQueue,
-            ILogger log)
+            ILogger log,
+            CancellationToken cancelToken)
         {
-            log.LogInformation("Processing message {SubmissionId}, stage {Stage}", workflowMessage.SubmissionId, workflowMessage.Stage);
+            Action<WorkflowQueueMessage> cancelAsync = (msg) =>
+            {
+                log.LogError("Timeout during workflow action execution.SubmissionId: { msg.SubmissionId}, Stage: { msg.Stage}", workflowMessage.SubmissionId, workflowMessage.Stage);
```
conniey: could condense this into:
```cs
var errorMessage = "Timeout during workflow action execution.SubmissionId: {msg.SubmissionId}, Stage: {msg.Stage}";

log.LogError(errorMessage, workflowMessage.SubmissionId, workflowMessage.Stage);
throw new TimeoutException(errorMessage); 

```

src/backend/Functions/ProcessWorkflowQueue.cs
```diff
@@ -14,17 +15,43 @@ public static class ProcessWorkflowQueue
...
+                    {
+                        //TODO: Add notification to client that analysis workflow has completed successfully
+                    }
+                }
+            }
+            else
             {
-                log.LogInformation("Queueing new message {SubmissionId}, stage {Stage}", workflowMessage.SubmissionId, nextMsg.Stage);
-                workflowMessageQueue.Add(nextMsg);
+                log.LogError("Timeout before workflow action started.SubmissionId: { msg.SubmissionId}, Stage: { msg.Stage}", workflowMessage.SubmissionId, workflowMessage.Stage);
```
conniey: potentially reuse that `errorMessage` down here.

src/backend/WorkflowManagement/WorkflowManager.cs
```diff
@@ -76,10 +77,10 @@ public static WorkflowQueueMessage GetFirstStage(string submissionId)
         /// Executes the action for the current stage in the workflow. After the action completes,
         /// returns a message for the next stage in the queue. 
         /// </summary>
-        public async Task<WorkflowQueueMessage> ExecuteActionsToNextStage(WorkflowQueueMessage currentMsg)
+        public async Task<WorkflowQueueMessage> ExecuteActionsToNextStage(WorkflowQueueMessage currentMsg, CancellationToken cancelToken)
         {
             //Execute the action
-            WorkflowStage nextStage = await actions[(int)currentMsg.Stage].ExecuteAsync(currentMsg.SubmissionId);
+            WorkflowStage nextStage = await actions[(int)currentMsg.Stage].ExecuteAsync(currentMsg.SubmissionId, cancelToken);
```
conniey: Since this could is a library, I would add `.ConfigureAwait(false)` if resuming on the same context does not matter.
```cs
WorkflowStage nextStage = await actions[(int)currentMsg.Stage].ExecuteAsync(currentMsg.SubmissionId, cancelToken).ConfigureAwait(false); 
```

src/backend/Functions/ProcessWorkflowQueue.cs
twsouthwick: ILogger is using structured logging, so the strings are actually different.

twsouthwick: I would think you should just have `{SubmissionId}` etc in the log string. Is there a reason you include the `msg`?

src/backend/Functions/ProcessWorkflowQueue.cs
```diff
@@ -14,17 +15,43 @@ public static class ProcessWorkflowQueue
...
-            log.LogInformation("Processing message {SubmissionId}, stage {Stage}", workflowMessage.SubmissionId, workflowMessage.Stage);
+            Action<WorkflowQueueMessage> cancelAsync = (msg) =>
+            {
+                log.LogError("Timeout during workflow action execution.SubmissionId: { msg.SubmissionId}, Stage: { msg.Stage}", workflowMessage.SubmissionId, workflowMessage.Stage);
+                throw new TimeoutException($"Timeout during workflow action execution. SubmissionId: {msg.SubmissionId} Stage: {msg.Stage}.");
+            };
+
+            //This cancellation token will let us know if this function has timed out.  When it times out, throw an exception so that the 
+            //function completes with a failure.  This will ensure that the message is removed from apiportworkflowqueue and automatically added to apiportworkflowqueue-poison so that
+            //we can treat a function timeout as an ordinary failure where the end user must resubmit their submission for analysis.
```
twsouthwick: Instead of the `if/else` you can call `cancelToken.ThrowIfCancelRequested()` and it will throw if needed.

src/backend/WorkflowManagement/WorkflowManager.cs
twsouthwick: Do you need any bounds checking here? Such as `actions.TryGetValue` or something?

src/backend/Functions/ProcessWorkflowQueue.cs
nicolehaugen79: Yes, leaving as-is for reason that taylor mentioned.

conniey: Good to know! Now I know. :D

nicolehaugen79: copy paste error...so I removed "msg" now :)

src/backend/Functions/ProcessWorkflowQueue.cs
nicolehaugen79: I didn't do this because ThrowIfCancellation doesn't allow me to pass my own message; I wanted the exception to include the submission id and stage for being able to query easily in AI.  Also, I want to ensure LogError can be called.

src/backend/Functions/ProcessWorkflowPoisonQueue.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Azure.WebJobs;
+using Microsoft.Extensions.Logging;
+using WorkflowManagement;
+
+namespace Functions
```
conniey: @nicolehaugen79 Since this is merged, you may have to open a PR to pull this into `PortabilityService.Functions/` rather than `Functions/`

src/backend/PortabilityService.Functions/Analyze.cs
```diff
@@ -24,40 +24,41 @@ public static class Analyze
...
             }
 
-            var submissionId = Guid.NewGuid().ToString();
-            log.LogInformation("Created submission id {SubmissionId}", submissionId);
-
             try
             {
+                log.LogInformation("Created submission id {SubmissionId}", submissionId);
+
                 var saved = await storage.SaveToBlobAsync(analyzeRequest, submissionId);
```
conniey: I don't think the comments about `await` with `.ConfigureAwait(false)` were addressed. :(

src/backend/WorkflowManagement/WorkflowManager.cs
nicolehaugen79: Sorry I missed it...done.

nicolehaugen79: Yes, I've been meaning to improve that...thanks for bringing it up.

PR #605, opened 2018-04-06
Rename projects to be consistent

jeremymeng:
While at it, Remove and Sort usings in touched files with the option of
`dotnet_sort_system_directives_first = false`.


jeremymeng:
Thanks @mjrousos and @twsouthwick

PR #610, opened 2018-04-13
code review feedback

nicolehaugen79:
Includes:

ConfigureAwait(false)

Debug.Asserts

Moved the poison workflow function to new project.


conniey:
@nicolehaugen79 This solution doesn't build for me. :(

* One of the errors is from the `cancelToken.ThrowIfCancellationRequested()`. Taylor had mentioned to add that in so you didn't have to provide the entire `if (!cancelToken.IsCancellationRequested) { } else { }` but do:
```csharp
cancelToken.ThrowIfCancellationRequested();

using (var ctr = cancelToken.Register(() => cancelAsync(workflowMessage)))
{
   //rest of logic
}
```

* `.ConfigureAwait(false)` should be paired with all the instances where you use `await` (I remember there being more but didn't mention them all. I can look through if you want). I think we'll probably add a Roslyn analyzer that can check for that.

conniey:
@dotnet-bot test this please

conniey:
@dotnet-bot test this please

nicolehaugen79:
@conniey - geesh, I should not have submitted my PR for review on the way out the door on a Friday. :(  Anyway...I removed the compile error from the rogue ThrowIfCancellationRequested line...this shouldn't have been there at all (see more info in my response to Taylor's comment).

Also, good catch on the ConfigureAwait...I was reading up on this since I haven't used it before, and missed that point.  It would be great to get a Roslyn check added for this.

PR #616, opened 2018-04-19
Mock functions for retrieving reports

chlowell:
Here are a couple of functions I needed for the client refactoring I'm working on. `Report` retrieves reports, `ReportFormats` lists available report formats. I included a binary for ApiPort to analyze and portability reports for it in json and Excel formats, generated by the deployed service. No tests included because these functions don't do much and aren't necessarily intended to be permanent--they just help mock the end-to-end scenario for the client.

This the design I settled on for the client to retrieve reports:
1. the client posts an `AnalyzeRequest` to the `Analyze` function
1. the `Analyze` function returns an `AnalyzeResponse` (not part of this PR)
   * `AnalyzeResponse` includes a report URL and auth token
1. the client polls the report URL (not part of this PR)
   * specifically it gets the report URL, providing the token in the Authorization header
1. the `Report` function responds with the report format specified by the request's Accept header


src/backend/PortabilityService.Functions/Report.cs
```diff
@@ -0,0 +1,94 @@
...
+                case "application/json":
+                    return JsonReport(req);
+                case "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet":
+                    return ExcelReport(req);
+                default:
+                    return req.CreateResponse(HttpStatusCode.UnsupportedMediaType);
+            }
+        }
+
+        public static bool ValidAccessKey(HttpRequestMessage request)
```
conniey: As long as this is temporary, can we put a TODO to replace with authorization or investigate?  I think we have a user story for it.. not sure.

src/backend/PortabilityService.Functions/Report.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            // simulate report generation taking some time
+            if (new Random().Next(10) < 4)
+            {
+                return req.CreateResponse(HttpStatusCode.NotFound);
+            }
+
+            switch (req.Headers.Accept.ToString())
+            {
+                case "application/json":
```
conniey: Hmm... Do you know if there is a constants class with these mime types? I see it being used here and in ReportFormat.cs and wonder if they can be consolidated.

chlowell: To my knowledge, no. I added one we can use for now.

PR #617, opened 2018-04-20
Delete breaking change documentation

chlowell:
This is maintained in [Microsoft/dotnet](https://github.com/Microsoft/dotnet/tree/master/Documentation/compatibility).


conniey:
Have you removed this from PortabilityTools.sln as well? IIRC, we have these documents linked in our solution.

chlowell:
Good catch, `Microsoft.Fx.Portability.Offline` embeds them all. Which raises the question, Are we removing offline breaking change analysis?

chlowell:
Closing this because we first must rework the `ApiPort.Offline` build to embed docs from Microsoft/dotnet.

PR #618, opened 2018-04-20
Fixing build issues with msbuild properties

conniey:
Fixing the properties for our service build by not signing.


conniey:
@chlowell I pushed an update to the build. It should work now.

PR #621, opened 2018-04-23
Add PortabilityService configuration provider

JJVertical:
Adding a configuration source that allows ASPNet Core to add configuration through a configuration builder.


conniey:
@JJVertical There appears to be a test failure in your [Tests](https://ci2.dot.net/job/Microsoft_dotnet-apiport/job/service/job/windows_nt_debug_prtest/13/console)

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Threading.Tasks;
+
+namespace PortabilityService.ConfigurationSource
+{
+    public class PortabilityServiceConfigurationSource : MemoryConfigurationSource
+    {
+        private const string DefaultConfigurationServiceUrl = "http://localhost:1443";
+        internal const string PortabilityServiceConfigurationRoot = "PortabilityServiceSettings";
+
+        public PortabilityServiceConfigurationSource(string configurationSection = PortabilityServiceConfigurationRoot)
```
conniey: I would remove this override. I'd rather people have to set an environment variable in the override below.

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
```diff
@@ -0,0 +1,89 @@
...
+            InitialData = GetConfigurationSectionSettings(configurationSection ?? throw new ArgumentNullException(nameof(configurationSection)), httpClient ?? throw new ArgumentNullException(nameof(httpClient))).GetAwaiter().GetResult();
+        }
+
+        /// <summary>
+        /// Gets the Portability configuration service URL from the environment
+        /// </summary>
+        /// <returns></returns>
+        private static Uri GetConfigurationServiceUrlFromEnvironment()
+        {
+            var urlFromEnvironment = Environment.GetEnvironmentVariable("PortabilityConfigurationServiceUrl");
```
conniey: I would put this in a constants class above.

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
```diff
@@ -0,0 +1,89 @@
...
+        /// <summary>
+        /// Creates an HTTPClient with the appsettings.json Url
+        /// </summary>
+        private static HttpClient CreateHttpClient(Uri configurationServiceUrl) => new HttpClient() { BaseAddress = configurationServiceUrl };
+
+        /// <summary>
+        /// Gets the configuration settings for the passed in section
+        /// </summary>
+        private static async Task<IEnumerable<KeyValuePair<string, string>>> GetConfigurationSectionSettings(string configurationSection, HttpClient httpClient)
+            => await GetJsonDataAsync<IEnumerable<KeyValuePair<string, string>>>($"/api/Configuration/sectionsettingslist/{configurationSection}", httpClient);
```
conniey: I would put this into a `Constants.cs` class.

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        public PortabilityServiceConfigurationSource(string configurationSection = PortabilityServiceConfigurationRoot)
+            : this(GetConfigurationServiceUrlFromEnvironment() ?? new Uri(DefaultConfigurationServiceUrl), configurationSection) { }
+
+        public PortabilityServiceConfigurationSource(Uri configurationServiceUrl, string configurationSection = PortabilityServiceConfigurationRoot)
+            : this(CreateHttpClient(configurationServiceUrl ?? throw new ArgumentNullException(nameof(configurationServiceUrl))), configurationSection ?? throw new ArgumentNullException(nameof(configurationSection))) { }
+
+        public PortabilityServiceConfigurationSource(HttpClient httpClient, string configurationSection = PortabilityServiceConfigurationRoot)
+        {
+            InitialData = GetConfigurationSectionSettings(configurationSection ?? throw new ArgumentNullException(nameof(configurationSection)), httpClient ?? throw new ArgumentNullException(nameof(httpClient))).GetAwaiter().GetResult();
```
conniey: nit: Separate into new lines to make this easier to follow:

```csharp
InitialData = GetConfigurationSectionSettings(
    configurationSection ?? throw new ArgumentNullException(nameof(configurationSection)),
    httpClient ?? throw new ArgumentNullException(nameof(httpClient))
)
.GetAwaiter().GetResult();
```

I would replace the initialization with an AsyncLazy<T>... Typically, you want your constructors to return quickly and async methods can block for an indeterminate period of time while you're waiting for the result.
https://blog.stephencleary.com/2013/01/async-oop-2-constructors.html

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        /// <summary>
+        /// Returns Json from the WebAPI call with passed in URI
+        /// </summary>
+        private static async Task<T> GetJsonDataAsync<T>(string uri, HttpClient httpClient)
+        {
+            using (var request = new HttpRequestMessage(HttpMethod.Get, uri))
+            {
+                request.Headers.Add("Accept", "application/json");
+                return await GetJsonDataAsync<T>(request, httpClient);
```
conniey: Add: `.ConfigureAwait(false)`

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
```diff
@@ -0,0 +1,89 @@
...
+        private static async Task<T> GetJsonDataAsync<T>(HttpRequestMessage request, HttpClient httpClient)
+        {
+            using (var response = await httpClient.SendAsync(request))
+            {
+                if (!response.IsSuccessStatusCode)
+                {
+                    return default;
+                }
+
+                var content = await response.Content.ReadAsStringAsync();
```
conniey: Add: `.ConfigureAwait(false)`

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
mjrousos: Agreed on not waiting for initialization in the ctor. Unfortunately, I don't know a good way to make ASP.NET Core configuration providers/sources async. One option might be to just store the URL and section name here and wait on retrieving the settings in a custom configuration provider's `Load` method. It's still not async, but at least the synchronous blocking is not in the ctor. This is what first-party providers like the [KeyVault config source/provider](https://github.com/aspnet/Configuration/tree/dev/src/Config.AzureKeyVault) do.

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
mjrousos: This should probably be in a config provider (rather than a source) to enable reloading, etc.

src/backend/PortabilityService.ConfigurationSource/PortabilityService.ConfigurationSource.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
```
mjrousos: netstandard2.0, right?

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
mjrousos: Also, have we tested whether the /sectionsettingslist endpoint properly handle arrays of settings? For example, with [this sample json configuration](https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/configuration/index/sample/ConfigJson/appsettings.json) will we correctly receive both wizards? It seems worth testing since the idea of [] config values may not map properly to an `IEnumerable<KVP>`. 

This will be important for PortabilityService since Serilog sinks are provided as an [] of settings, as well as Ocelot routing information (though the latter might not be in shared configuration).

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
JJVertical: I didn't do it this way because I didn't want null passed in for the URI

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
JJVertical: Added to constants class

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
```diff
@@ -0,0 +1,89 @@
...
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace PortabilityService.ConfigurationSource
+{
+    public class PortabilityServiceConfigurationSource : MemoryConfigurationSource
+    {
+        private const string DefaultConfigurationServiceUrl = "http://localhost:1443";
```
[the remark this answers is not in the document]
JJVertical: Moved this to constants.cs file. I like having the hard-coded URL because this is what the configurationservice is configured for when doing local development. This should come from the environment outside of local development.

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
JJVertical: Added

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
JJVertical: Added

src/backend/PortabilityService.ConfigurationSource/PortabilityService.ConfigurationSource.csproj
JJVertical: Updated this to netstandard

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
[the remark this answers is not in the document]
conniey: When we startup the configuration service, it should be setting the environment variable, `PortabilityConfigurationServiceUrl`.

If we want to ensure that all our consuming projects have the environment variable defined, we can move that logic into our `Directory.Build.Props` or `Directory.Build.targets` for development.

The downside to having this logic compiled into a library, is that if we decide to change the localhost:1443 to like localhost:1000 in the service, we will be forced to recompile that library, find the string in another location and replace that.

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
conniey: Sounds good.

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
JJVertical: We have a test that checks that the sectionsettingslist endpoint handles arrays in the ConfigurationService tests.

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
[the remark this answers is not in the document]
conniey: I was thinking more about this, and msbuild doesn't allow by default to set Environment Variables from your msbuild script. I think it makes sense because you wouldn't want to run MSBuild and have it modify your user environment.

I was wondering if it is better to have something like [PortabilityServiceConfigurationSettings.Development.json](https://github.com/Microsoft/dotnet-apiport/blob/service/src/backend/PortabilityService.ConfigurationService/PortabilityServiceConfigurationSettings.Development.json) in the root of our repository and import them like we do in [Directory.Build.targets](https://github.com/Microsoft/dotnet-apiport/blob/service/Directory.Build.targets#L37-L39) where we include it in an itemgroup and use it only for development.

This would remove the entire use of a environment variable.

What do you think @twsouthwick?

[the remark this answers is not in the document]
twsouthwick: This is tricky as it's a bootstrapping problem for configuration... Which aspects here are potentially changeable? The host (`localhost`) and/or port (`1443`)? What about conditioning it off of `IHostingEnvironment.IsDevelopment()`?

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
twsouthwick: We should add the fxcop analyzers that check for this (not in this PR, but could be a task or something)

src/backend/PortabilityService.ConfigurationSource/ConfigurationSource.cs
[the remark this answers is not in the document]
JJVertical: Updated to remove the constant and throw if environment can't be set as discussed off line

src/backend/PortabilityService.ConfigurationProvider/PortabilityService.ConfigurationProvider.csproj
```diff
@@ -0,0 +1,28 @@
...
+      <DesignTime>True</DesignTime>
+      <AutoGen>True</AutoGen>
+      <DependentUpon>Resources.resx</DependentUpon>
+    </Compile>
+  </ItemGroup>
+
+  <ItemGroup>
+    <EmbeddedResource Update="Resources\Resources.resx">
+      <Generator>ResXFileCodeGenerator</Generator>
+      <LastGenOutput>Resources.Designer.cs</LastGenOutput>
```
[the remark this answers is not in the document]
JJVertical: I used this in the ConfigurationService, but used the Designer.cs file here to be consistent with other libraries

src/backend/PortabilityService.ConfigurationProvider/PortabilityServiceConfigurationProvider.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// <summary>
+        /// Returns Json from the WebApi call with the passed in request message
+        /// </summary>
+        private async Task<T> GetJsonDataAsync<T>(HttpRequestMessage request)
+        {
+            var response = await _httpClient.SendAsync(request);
+
+            if (!response.IsSuccessStatusCode)
+            {
+                return default;
```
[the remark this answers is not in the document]
JJVertical: I will create a separate pull request for the logging from the provider

PR #624, opened 2018-04-25
Remove ApiPort.VisualStudio.2015

chlowell:
This removes our Visual Studio 2015 extension.


chlowell:
Sorry to report it doesn't. Resolving the binding conflicts by upgrading to newer dependencies introduces ~40 warnings from VS SDK analyzers, which I haven't worked through.

conniey:
> Sorry to report it doesn't. Resolving the binding conflicts by upgrading to newer dependencies introduces ~40 warnings from VS SDK analyzers, which I haven't worked through.

Maybe we shouldn't upgrade them. Ahaha.

chlowell:
I'm thinking the same thing but haven't determined exactly which dependencies we need. We can hold off merging this until that's done, this work isn't really complete if we ship assemblies we don't use.

chlowell:
I used a newer version of the VS threading library which does not include the analyzers 🙈 and eliminates the binding warnings. Two birds, one hug.

PR #632, opened 2018-05-02
Update Gateway to use PortabilityService ConfigurationProvider

JJVertical:


tests/backend/PortabilityService.Gateway.Tests/AspNetCoreTestHelpers/WebHostHelper.cs
```diff
@@ -15,12 +19,23 @@ internal static class TestGateway
...
+
+        private static IWebHostBuilder CreateTestWebHost() =>
+            WebHost.CreateDefaultBuilder()
+            .ConfigureAppConfiguration(ConfigureTestAppConfiguration)
+            .UseStartup<Startup>();
+
+        private static void ConfigureTestAppConfiguration(WebHostBuilderContext context, IConfigurationBuilder configBuilder)
+        {
+            configBuilder.AddJsonFile("reroutes.json");
+            configBuilder.AddPortabilityServiceConfiguration(new Uri("http://localhost"), ConfigurationProviderConstants.PortabilityServiceConfigurationRoot, true);
```
mjrousos: Should we maybe just leave this out for test scenarios since it was causing problems and we don't need configuration from the service to test the gateway?

tests/backend/PortabilityService.Gateway.Tests/PortabilityService.Gateway.Tests.csproj
```diff
@@ -15,6 +15,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\..\src\backend\PortabilityService.ConfigurationProvider\PortabilityService.ConfigurationProvider.csproj" />
```
mjrousos: As mentioned before, I think the tests don't even need this.

src/backend/PortabilityService.Gateway/Program.cs
```diff
@@ -52,7 +54,7 @@ private static void ConfigureLogging(WebHostBuilderContext context, ILoggingBuil
 
             // Create Serilog configuration from app configuration
             var serilogLoggerConfiguration = new LoggerConfiguration()
-                .ReadFrom.Configuration(config);
+                .ReadFrom.Configuration(config.GetSection(ConfigurationProviderConstants.PortabilityServiceConfigurationRoot));
```
mjrousos: It would be better to just read from the configuration directly (rather than a config service-specific section) so that if we end up falling back to appsettings.json (if the config service is down or something) everything just works.

tests/backend/PortabilityService.Gateway.Tests/AspNetCoreTestHelpers/WebHostHelper.cs
JJVertical: Removed

tests/backend/PortabilityService.Gateway.Tests/PortabilityService.Gateway.Tests.csproj
JJVertical: Removed this also

src/backend/PortabilityService.Gateway/Program.cs
JJVertical: Updated the ConfigService to return the section settings without nulls and settings. Now this will just be Config instead of Config.GetSection

PR #636, opened 2018-05-08
Remove null values and passed in section from ConfigService result

JJVertical:
Change the ConfigurationService API calls to remove null configuration values and the SectionName passed in from the result.

### DIFF
--- a/src/backend/PortabilityService.ConfigurationService/Controllers/ConfigurationController.cs
+++ b/src/backend/PortabilityService.ConfigurationService/Controllers/ConfigurationController.cs
@@ -8,6 +8,7 @@ using Microsoft.Extensions.Localization;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
 using System;
+using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


src/backend/PortabilityService.ConfigurationService/Controllers/ConfigurationController.cs
```diff
@@ -90,7 +87,11 @@ public ObjectResult GetSectionSettingsList(string sectionName)
             }
 
             _logger.LogInformation(LogMessageFormat, nameof(GetSectionSettingsList), _localizer["ReturningSectionSettings", sectionName].Value);
-            return Ok(_configuration.GetSection(sectionName).AsEnumerable());
+            return Ok(_configuration.GetSection(sectionName)
+                      .AsEnumerable()
+                      .Where(v => v.Value != null)
+                      .Select(k => new KeyValuePair<string, string>(k.Key.Substring(_baseConfigSectionName.Length + 1), k.Value))
+                      .DefaultIfEmpty());
```
mjrousos: Why `DefaultIfEmpty`?

PR #638, opened 2018-05-09
Fix coding style warnings

jeremymeng:


src/backend/PortabilityService.AnalysisEngine/Startup.cs
```diff
@@ -35,7 +35,7 @@ public void ConfigureServices(IServiceCollection services)
         }
 
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
-        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
+        public static void Configure(IApplicationBuilder app, IHostingEnvironment env)
```
mjrousos: I've always left this as an instance method even when it could be static just so all my `Startup.Configure` methods match. I don't feel strongly about it, though.

jeremymeng: Agree. I considered other options too
* suppress the warning for the whole solution
* suppress the warning for this project
* suppress the warning for this particular place
but thought this was the least ugly one.

@twsouthwick mentioned that he will disable most of the CA warnings that do make a lot of sense.  I am going to abandon this PR for now.

twsouthwick: See #642 - I've disabled this rule. FYI this was showing up as a warning in the dev environment and on release builds warnings are treated as errors. Not sure why the PR checks weren't getting them, but they should show up in VS

PR #647, opened 2018-05-15
Remove docker-compose project from the sln

mjrousos:
It was blocking solution-level restore from working (dotnet/cli#6178) and didn't work to launch Docker containers from VS anyhow (Microsoft/DockerTools#54)


mjrousos:
Probably don't need it. I left it in case we want to add it back to the sln later, but it's not doing anything at the moment.

PR #650, opened 2018-05-17
Merge dev into service

twsouthwick:


twsouthwick:
This needs to be merged, not squashed, so that we can maintain branch history correctly. Once it's approved, I'll enable that form of merge and handle it.

PR #652, opened 2018-05-17
Service Fabric Mesh templates

mjrousos:
This adds SF Mesh templates (and deployment instructions) for PortabilityService gateway and config service.


deploy/PortabilityService.ServiceFabricMeshApplication/Reference/Deploy.md
```diff
@@ -0,0 +1,16 @@
+Manual SeaBreeze Deployment Instructions
```
twsouthwick: Should this be "mesh"?

deploy/PortabilityService.ServiceFabricMeshApplication/PortabilityService-1709.json
```diff
@@ -0,0 +1,133 @@
...
+    "location": {
+      "type": "string",
+      "defaultValue": "eastus",
+      "metadata": {
+        "description": "Location of the resources."
+      }
+    },
+    "registry-server": {
+      "type": "string",
+      "defaultValue": "portabilityservice.azurecr.io",
```
twsouthwick: Should this be public?

deploy/PortabilityService.ServiceFabricMeshApplication/Reference/Deploy.md
mjrousos: Good catch; forgot to update this one :)

deploy/PortabilityService.ServiceFabricMeshApplication/PortabilityService-1709.json
mjrousos: Good question. I don't think there's much danger including the registry name and user name as defaults since the password still needs to be provided. It may be better to not include them, though, just because they aren't useful to anyone without access to this particular registry.

PR #653, opened 2018-05-17
Disable VSIX deployment during CI build

twsouthwick:
This can potentially add quite a while to the build process and we don't need it to actually be deployed

### DIFF
--- a/.vsts.ci.yaml
+++ b/.vsts.ci.yaml
@@ -54,7 +54,7 @@ steps:
   inputs:
     solution: PortabilityTools.sln
     vsVersion: 15.0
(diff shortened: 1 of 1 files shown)


.vsts.ci.yaml
```diff
@@ -63,7 +63,7 @@ steps:
   inputs:
     solution: PortabilityTools.sln
     vsVersion: 15.0
-    msbuildArgs: '/t:build /bl:$(Build.SourcesDirectory)\bin\$(BuildConfiguration)\msbuild.binlog'
+    msbuildArgs: '/t:build /p:DeployExtension=false /bl:$(Build.SourcesDirectory)\bin\$(BuildConfiguration)\msbuild.binlog'
```
chlowell: Can you add `/m` here so we get a parallel build?

twsouthwick: Done

PR #655, opened 2018-05-17
Add deployment artifacts for Analysis Engine service

jeremymeng:


src/backend/PortabilityService.AnalysisEngine/Dockerfile
```diff
@@ -0,0 +1,16 @@
+### Build Stage
+FROM microsoft/aspnetcore-build:2.0.7-2.1.105 AS build
+WORKDIR /src
+
+COPY NuGet.config .
+COPY ./src .
+WORKDIR /src/backend/PortabilityService.AnalysisEngine
+RUN dotnet publish PortabilityService.AnalysisEngine.csproj -c Release -o /app
+
```
jeremymeng: @mjrousos It looks like we just need 'publish' command which will automatically restore and build before publishing. Is there benefit to have separate steps (better for debugging issue maybe)?

mjrousos: Separating restore and build is beneficial because that way the output of each step will be a separate image layer. Then, in the future, if a change is made that requires a re-build (but not a re-restore), `docker build` can re-use the layer containing restore output and not have to spend time restoring NuGet packages again.

In our case, I think separating build and publish probably doesn't really matter (I just copied the Dockerfile format from a template).

docker-compose.yml
```diff
@@ -13,6 +13,18 @@ services:
     environment:
       "PortabilityConfigurationServiceUrl": "http://portabilityservice.configurationservice"
 
+  portabilityservice.analysisengine:
+    image: ${DOCKER_REGISTRY}portabilityservice-analysisengine:${PORTABILITYSERVICE_VERSION:-latest}
+    build:
+      context: .
+      dockerfile: src/backend/PortabilityService.AnalysisEngine/Dockerfile
+    ports:
+      - "80:80"
```
mjrousos: I think this should just be "80" (instead of "80:80") since we don't need to expose the analysis service externally (it will be called from the gateway).

jeremymeng: Fixed.  Yes I agree.

jeremymeng:
I've updated the Docker file to be consistent with other existing ones.

mjrousos:
LGTM

PR #659, opened 2018-05-21
Give SF services unique ports

mjrousos:
Make sure each service has a unique port

PortabilityService services locate each other in SF deployments
with SF DNS. Manually assign unique ports since SF DNS doesn't
support dynamic ports and SF Reverse Proxy isn't an option since it
doesn't work in Linux clusters yet.

https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-dnsservice
https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-reverseproxy#communicating-by-using-the-reverse-proxy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ src/Microsoft.Fx.Portability.Offline/data/
 *.nuget.props
 *.nuget.targets
 .vscode
+.data/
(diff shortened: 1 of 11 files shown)


jeremymeng:
LGTM.

Possibly in a future change we can pass in the docker image versions, or even the tags as parameters to the manifests.

PR #661, opened 2018-05-21
Update Portability Service Gateway routes

mjrousos:


deploy/PortabilityService.ServiceFabricApplication/DeploySF.ps1
```diff
@@ -131,6 +137,10 @@ if ($EnvironmentName)
 {
     $parameters.EnvironmentName = $EnvironmentName
 }
+if ($ImageTag)
+{
+    $parameters.ImageTag = $ImageTag
```
twsouthwick: What's the point of this if the tags cannot be parameterized?

mjrousos: Just added for future use. Though researching, there may not be an easy way to use it so I'll remove it for now.

mjrousos:
SF doesn't allow the tags to be parameters so we'll have to do some token substitution ourselves.

conniey:
@mjrousos Is this ready to be merged?

twsouthwick:
Should we file a feature request to allow that to be parametized? It would make deployment much easier

PR #664, opened 2018-05-24
Generate reports with an Azure Function

chlowell:
This replaces the Function I added earlier to return mock reports with one that generates and returns real reports, albeit for hard-coded data because the analysis engine doesn't yet analyze anything. This completes the end-to-end scenario for ApiPort. I also updated the `ReportFormat` Function to return the formats we have writers for.

I learned along the way that generating a report requires an `AnalyzeRequest` and `AnalyzeResponse` both; that might be worth refactoring in the future. For now, we'll have to hang on to `AnalyzeRequest`s.


src/backend/PortabilityService.AnalysisEngine/Controllers/AnalyzeController.cs
```diff
@@ -89,25 +83,11 @@ private Task<AnalyzeResult> AnalyzeRequestAsync(AnalyzeRequest analyzeRequest, s
                 //TODO: invoke the real analysis engine to do the work
                 //return _requestAnalyzer.AnalyzeRequest(analyzeRequest, submissionId);
```
twsouthwick: Probably would be better to encapsulate this in a dummy _requestAnalyzer and register that so this code won't have to change later. Include a comment that the embedded resource itself should be removed if the code is removed

src/backend/PortabilityService.Functions/Report.cs
```diff
@@ -3,100 +3,115 @@
...
+            if (analyzeRequest == null)
             {
-                // TODO this should return 202 only when the analyze request is known to have been received,
-                // 404 otherwise (e.g. when the client tries to get the report before its analyze request
-                // has propagated)
-                var response = req.CreateResponse(HttpStatusCode.Accepted);
-                response.Headers.RetryAfter = new RetryConditionHeaderValue(TimeSpan.FromSeconds(2d));
+                // the request may have been received but not yet stored => the client should poll again
+                var res = req.CreateResponse(HttpStatusCode.NotFound);
+                res.Headers.RetryAfter = new RetryConditionHeaderValue(TimeSpan.FromSeconds(4d));
```
twsouthwick: why the `d`?

src/backend/PortabilityService.Functions/Report.cs
```diff
@@ -3,100 +3,115 @@
...
 
-                return response;
+                return res;
             }
 
-            switch (req.Headers.Accept.ToString())
+            var response = req.CreateResponse();
+            try
+            {
+                response.Content = GetReportContent(req.Headers.Accept.SingleOrDefault(), analyzeRequest, analyzeResult);
```
twsouthwick: What happens if no accept headers are sent? This will throw

src/backend/PortabilityService.Functions/Report.cs
```diff
@@ -3,100 +3,115 @@
...
         {
-            // TODO generate and persist a new unique key in Analyze, validate it here
-            var authHeader = request.Headers.Authorization;
-            if (authHeader == null || !authHeader.Scheme.Equals("Bearer", StringComparison.Ordinal))
+            AnalyzeRequest analyzeRequest = null;
+            AnalyzeResult analyzeResult = null;
+            try
             {
-                return false;
+                analyzeRequest = await storage.RetrieveRequestAsync(submissionId);
```
twsouthwick: Can you move the variable declarations and definitions together? You can include the `return (analyzeRequest, analyzeResult)` in the try statement

src/backend/PortabilityService.Functions/Report.cs
```diff
@@ -3,100 +3,115 @@
...
-            var chars = submissionId.ToCharArray();
-            Array.Reverse(chars);
-            var expectedToken = new string(chars);
-
-            return token.Equals(expectedToken, StringComparison.Ordinal);
+            return (analyzeRequest, analyzeResult);
         }
 
-        public static HttpResponseMessage JsonReport(HttpRequestMessage request)
+        private static readonly ITargetMapper s_targetMapper = new TargetMapper();
```
twsouthwick: Can we use DI for this?

src/backend/PortabilityService.Functions/Report.cs
```diff
@@ -3,100 +3,115 @@
...
+        public static IReadOnlyCollection<IReportWriter> ReportWriters { get; } = new IReportWriter[]
```
twsouthwick: Can we use DI for this?

src/backend/PortabilityService.Functions/ReportFormat.cs
```diff
@@ -29,20 +28,8 @@ public static class ReportFormat
...
-                FileExtension = ".xlsx"
-            },
-            new ResultFormatInformation
-            {
-                DisplayName = "Json",
-                FileExtension = ".json",
-                MimeType = "application/json"
-            }
-        };
+        private static ResultFormatInformation[] Formats { get; } = Report.ReportWriters
```
twsouthwick: DI?

src/backend/PortabilityService.WorkflowManagement/WorkflowManager.cs
```diff
@@ -84,7 +81,7 @@ public static WorkflowQueueMessage GetFirstStage(string submissionId)
         public async Task<WorkflowQueueMessage> ExecuteActionsToNextStage(WorkflowQueueMessage currentMsg, CancellationToken cancelToken)
         {
             //Get action corresponding to the current message's workflow stage
-            Debug.Assert(actions.Length - 1 > (int)currentMsg.Stage, "Stage must be within bounds of Actions array.");
+            Debug.Assert((int)currentMsg.Stage < actions.Length, "Stage must be within bounds of Actions array.");
```
twsouthwick: Should we have this throw instead of be debug? That way we can catch it in production if it ever occurs

src/lib/Microsoft.Fx.Portability.Azure/ReversedIdTokenValidator.cs
```diff
@@ -0,0 +1,29 @@
...
+            if (authHeader == null || !authHeader.Scheme.Equals("Bearer", StringComparison.Ordinal))
+            {
+                return false;
+            }
+
+            var token = authHeader.Parameter;
+            var submissionId = request.RequestUri.Segments.Last();
+            var chars = submissionId.ToCharArray();
+            Array.Reverse(chars);
+            var expectedToken = new string(chars);
```
twsouthwick: Probably not a big deal, but you can also just compare each character with a for loop to not allocate the string

tests/backend/PortabilityService.Functions.Tests/ReportFormatTests.cs
```diff
@@ -0,0 +1,56 @@
...
+        [Fact]
+        public void RespondsWithOneFormatForDefaultArg()
+        {
+            var request = new HttpRequestMessage { Method = HttpMethod.Get };
+            request.Properties.Add(HttpPropertyKeys.HttpConfigurationKey, new HttpConfiguration());
+
+            var response = ReportFormat.Run(request, "default");
+
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.True(response.TryGetContentValue(out ResultFormatInformation defaultFormat));
```
twsouthwick: Should there be any asserts on the `defaultFormat` result?

tests/backend/PortabilityService.Functions.Tests/ReportTests.cs
```diff
@@ -0,0 +1,84 @@
...
+        [Fact]
+        public async Task ReturnsUnauthorizedForInvalidToken()
+        {
+            var storage = Substitute.For<IStorage>();
+            storage.RetrieveRequestAsync(Arg.Any<string>()).Returns(new AnalyzeRequest());
+
+            var tokenValidator = Substitute.For<IReportTokenValidator>();
+            tokenValidator.RequestHasValidToken(Arg.Any<HttpRequestMessage>()).Returns(false);
+
+            var request = new HttpRequestMessage(HttpMethod.Get, "");
```
twsouthwick: dispose the request (applies to all methods in the file)

src/backend/PortabilityService.Functions/Report.cs
chlowell: I know it's unnecessary, I somehow developed the habit of being pedantic about literals like this.

src/backend/PortabilityService.Functions/Report.cs
chlowell: I mistakenly expected `SingleOrDefault` to be lazy and therefore not realized until after the null check. Also, turns out Postman sends `*/*` when Accept isn't specified. Today I learned.

src/backend/PortabilityService.WorkflowManagement/WorkflowManager.cs
chlowell: I think it makes sense as a debug assert because `currentMsg.Stage` is an enum. If the assertion fails, that's a bug. I'm also thinking we won't need, and so won't deploy, this code.

src/lib/Microsoft.Fx.Portability.Azure/ReversedIdTokenValidator.cs
chlowell: I see I lost the TODO about replacing this with a real thing, I'll add it back :suspect:

src/backend/PortabilityService.Functions/Report.cs
chlowell: I don't see a better way. If the declarations and return are in the `try` block, the outer scope must return something ugly like `(null, null)` and lose the capability of returning an instance of `analyzeRequest` when assigning `analyzeResult` throws, which would alter the Function's response. But maybe I'm just not seeing it--I'm grateful for any suggestions here.

twsouthwick: ah - I thought it was a finally or a catch-rethrow block. I'd still prefer the `(null,null)` as it makes it easier to reason about what is happening

src/backend/PortabilityService.Functions/ReportFormat.cs
twsouthwick: For me, it's more of consistency. I know DI isn't fully supported in functions, but I just want to make sure we're using the appropriate patterns where possible

src/backend/PortabilityService.Functions/Report.cs
```diff
@@ -3,100 +3,115 @@
...
+        {
+            new ExcelReportWriter(s_targetMapper),
+            // TODO HtmlReportWriter fails, possibly because RazorEngine can't write temp files to disk
+            //new HtmlReportWriter(s_targetMapper),
+            new JsonReportWriter()
+        };
+
+        public static ByteArrayContent GetReportContent(MediaTypeWithQualityHeaderValue mediaType, AnalyzeRequest analyzeRequest, AnalyzeResult analyzeResult)
```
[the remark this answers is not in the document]
chlowell: I intended to unit test them but decided to stick to testing `Run` alone because it's the client's interface.

PR #665, opened 2018-05-26
Use the real request analyzer to analyze requests

jeremymeng:


src/backend/PortabilityService.AnalysisService/IServiceSettings.cs
```diff
@@ -0,0 +1,86 @@
...
+    /// <summary>
+    /// Portability service settings
+    /// </summary>
+    public interface IServiceSettings
+    {
+        /// <summary>
+        /// How often to check for updates for data stored in Azure Storage.
+        /// Used to check for updates for catalog.bin and fxmember.json.gz
+        /// </summary>
+        TimeSpan UpdateFrequency { get; }
```
chlowell: This could be simpler--instead of updating the data while the service is live, we can redeploy the service. The orchestrator should make this cheap and orderly.

src/backend/PortabilityService.AnalysisService/appsettings.Development.json
```diff
@@ -0,0 +1,10 @@
+﻿{
+  "BlobStorageConnectionString": "UseDevelopmentStorage=true",
+  "CatalogStorageConnectionString": "UseDevelopmentStorage=true",
```
chlowell: If you put these in a `ConnectionStrings` object, you can use the more idiomatic `IConfiguration.GetConnectionString` in startup.

chlowell:
FYI I'm working on the failing test and it's unrelated to code changes. In our CI build the test class initialization isn't happening as expected.

src/backend/PortabilityService.AnalysisService/Dockerfile
```diff
@@ -0,0 +1,24 @@
+### Build Stage
+FROM microsoft/aspnetcore-build:2.0.7-2.1.105 AS build
```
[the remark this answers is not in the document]
jeremymeng: I thought about that but figured that it might be better to wait for RTM then upgrade our projects to 2.1 in a separate change.  We are not using the optional value anywhere.

src/backend/PortabilityService.AnalysisService/IServiceSettings.cs
jeremymeng: Agree. Though we probably should do similar Retry-After in the DataCatalog service client instead as this will replaced soon.

src/backend/PortabilityService.AnalysisService/appsettings.Development.json
jeremymeng: Good points

src/backend/PortabilityService.AnalysisService/DependencyInjection/DependencyModule.cs
```diff
@@ -0,0 +1,97 @@
...
+
+                return new CatalogInBlobIndexCache(settings, CancellationToken.None);
+            })
+                .As<IObjectCache<CatalogIndex>>()
+                .As<IObjectCache>()
+                .SingleInstance()
+                .OnActivated(o => o.Instance.Start())
+                .AutoActivate();
+
+            // TODO (yumeng): replace UnioningApiCatalogLookup with a concrete type implementing IApiCatalogLookup
```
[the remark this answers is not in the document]
jeremymeng: As mentioned above, One option is to use Retry-After for the catalog service.

PR #666, opened 2018-05-30
Fix failing test

chlowell:
`WorkflowManager`'s  initialization methods are not thread-safe. xUnit runs tests in parallel, and it's possible for test code to execute against `WorkflowManager` with its default initialization, which includes analysis actions which will attempt to send HTTP requests. These will fail in the test environment.


chlowell:
Another problem is that the `Analyze` function initializes `WorkflowManager`. So there's no easy fix here, we need to revisit this part of the design.

conniey:
@chlowell Can you create an issue for this so we revisit this problem? Or create a user story for it so it does not become lost?

chlowell:
I created a story to track this.

PR #676, opened 2018-06-08
Use the smaller size alpine docker images

jeremymeng:


jeremymeng:
Trying out the smaller alpine docker images to see whether there's any savings in docker build time.

jeremymeng:
The build errors indicate that the build agent is using Docker Windows Containers.

jeremymeng:
I did an experiment to disable the docker steps on VS2017 agent and add another build phase on Hosted Linux agent to just run `docker-compose`. It worked.  So looks like we can do multiple-agent build and move docker steps to run on hosted Linux agent.

PR #680, opened 2018-06-18
Added function to expose available targets

coderandhiker:
Exposes a list of available targets; this data is currently a static copy of the data from the legacy apiport /api/targets endpoint.


src/backend/PortabilityService.Functions/Targets.cs
```diff
@@ -0,0 +1,51 @@
...
+            var content = new StringContent(targets, Encoding.UTF8, "application/json");
+            return new HttpResponseMessage { StatusCode = System.Net.HttpStatusCode.OK, Content = content};
+        }
+
+        /// <summary>
+        /// This method is a placeholder, and will be replaced with an asynchronous call to fetch catalog data in a future sprint
+        /// </summary>
+        /// <returns>A Task<string> representing the async task that fetches the catalog of supported targets</string></returns>
+        /// <param name="log">logging provider instance</param>
+        private static async Task<string> GetStaticListOfTargets(ILogger log)
```
conniey: You can remove `async/await` and just return the Task. This removes the overhead of creating that entire async/await state machine if you're not doing anything with the Task after it completes inside this function.

src/backend/PortabilityService.Functions/Targets.cs
```diff
@@ -0,0 +1,51 @@
+﻿using Microsoft.Azure.WebJobs;
```
conniey: Do we happen to have any tests for this? Or the functions in general?

coderandhiker: Tests are under \test\backend\PortabilityService.Functions.Tests for those Functions that are doing more work than returning a static string

src/backend/PortabilityService.Functions/Targets.cs
conniey: nit: Naming `GetStaticListofTargetsAsync`

src/backend/PortabilityService.Functions/Targets.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// <returns>A Task<string> representing the async task that fetches the catalog of supported targets</string></returns>
+        /// <param name="log">logging provider instance</param>
+        private static Task<string> GetStaticListOfTargets(ILogger log)
+        {
+            return Task.Run(() =>
+            {
+                var targets = string.Empty;
+                try
+                {
+                    targets = Encoding.UTF8.GetString(Properties.Resources.ListOfTargets);
```
twsouthwick: you can just return this so as to not create an unnecessary local variable

tests/backend/PortabilityService.Functions.Tests/TargetsTests.cs
```diff
@@ -0,0 +1,33 @@
...
+            var tokenValidator = Substitute.For<IReportTokenValidator>();
+            tokenValidator.RequestHasValidToken(Arg.Any<HttpRequestMessage>()).Returns(false);
+
+            using (var request = new HttpRequestMessage(HttpMethod.Get, string.Empty))
+            {
+                var response = await Targets.Run(request, Substitute.For<ILogger>());
+                Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+                var contents = await response.Content.ReadAsStringAsync();
+#pragma warning disable CA1307 // Specify StringComparison (this string is static at this time)
+                Assert.Contains(".NET Core + Platform Extensions", contents);
```
twsouthwick: The StringComparison should be ordinal for compares of this sort

conniey:
@coderandhiker Looks like the CI build is failing because there should be some header, probably with the MIT information at the top of the file?

coderandhiker:
Rebuilt in VSTS - prior build failure was a transient network failure during a git fetch

PR #684, opened 2018-06-19
Combine catalog lookup implementation

twsouthwick:
There are two implementations of the lookup that can be combined as we are now in one repo.


src/lib/Microsoft.Fx.Portability/ObjectModel/CloudApiCatalogLookup.cs
```diff
@@ -65,121 +61,88 @@ public CloudApiCatalogLookup(DotNetCatalog catalog)
...
-            Dictionary<string, Version> targets;
             introducedVersion = null;
 
             // The docId is a member in the target if:
             //  - There is an entry for the API.
             //  - The entry for the API contains the target.
             //  - The version for when the API was introduced is before (or equal) to the target version.
-            if (!_apiMapping.TryGetValue(docId, out targets))
+            if (!_apiMapping.TryGetValue(docId, out var targets))
                 return false;
```
chlowell: It pains me to pick this nit, but `{ }`?

conniey:
Yay! Less random classes lying around! 👍

PR #687, opened 2018-06-19
Add breaking changes & recommended changes to AnalysisService

chlowell:
This adds breaking and recommended change data to the AnalysisService. The data is parsed from markdown on the local filesystem at configurable paths. I updated the Dockerfile to acquire the markdown at build so runtime images will launch with the data in place. I also added a readiness check with Kubernetes semantics so we have a mechanism to prevent requests flowing to the service before the data is deserialized.

### DIFF
--- a/src/backend/PortabilityService.AnalysisService/DependencyInjection/ApiPortDataModule.cs
+++ b/src/backend/PortabilityService.AnalysisService/DependencyInjection/ApiPortDataModule.cs
@@ -1,0 +1,97 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Autofac;
+using Microsoft.Fx.Portability;
+using Microsoft.Fx.Portability.ObjectModel;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace PortabilityService.AnalysisService
+{
+    public class ApiPortDataModule : Module
+    {
+        protected override void Load(ContainerBuilder builder)
+        {
+            builder.Register(LoadApiPortData).SingleInstance();
+            builder.RegisterType<GitCatalogRecommendations>().As<IApiRecommendations>().SingleInstance();
+        }
+
+        private ApiPortData LoadApiPortData(IComponentContext context)
+        {
+            var settings = context.Resolve<IServiceSettings>();
+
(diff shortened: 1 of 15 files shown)


src/lib/Microsoft.Fx.Portability.Git/ApiPortGitRepositoryStreamBuilder.cs
```diff
@@ -0,0 +1,356 @@
...
+using System.Linq;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace Microsoft.Fx.Portability
+{
+    /// <summary>
+    /// Builds <see cref="ApiPortData"/> using a Git repository stream.
+    /// </summary>
+    public class ApiPortGitRepositoryStreamBuilder
```
twsouthwick: Do we still use this? You're downloading the data directly, right? If so, this probably isn't needed

src/backend/PortabilityService.AnalysisService/DependencyInjection/ApiPortDataModule.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            builder.Register(LoadApiPortData).SingleInstance();
+            builder.RegisterType<GitCatalogRecommendations>().As<IApiRecommendations>().SingleInstance();
+        }
+
+        private ApiPortData LoadApiPortData(IComponentContext context)
+        {
+            var settings = context.Resolve<IServiceSettings>();
+
+            var breakingChanges = Directory.EnumerateFiles(settings.BreakingChangesPath, "*.md")
```
jeremymeng: Do we need to handle empty `settings.BreakingChangesPath` here, or let it bubbles up?

chlowell: I like letting the exception bubble up unhandled because I don't think there's a reasonable way to handle it and carry on; it signals a botched deployment.

chlowell: Also, note the readiness check won't return ok when no breaking changes are loaded, so the platform shouldn't route traffic to the service if the path is incorrect.

jeremymeng: Sounds good. A log might help but in this case it probably would be obvious what goes wrong.

PR #711, opened 2018-09-20
Fix build warning

conniey:
Surprised that our build broke since it should have failed on our gated CI release build.


twsouthwick:
Probably from changes that occurred concurrently and I don't think we require the PR being up-to-date

PR #712, opened 2018-09-20
More style cop rule fixes. Part 5

conniey:
Fixes:
* SA1500, SA1501, SA1503


src/ApiPort/ApiPort.VisualStudio.Common/Models/TargetPlatform.cs
```diff
@@ -104,36 +105,37 @@ public int CompareTo(TargetPlatform other)
             // information, like whether e1 comes before e2 (returning -1) or e1
             // comes after e2 (returning +1).
             using (var e1 = Versions.GetEnumerator())
-            using (var e2 = other.Versions.GetEnumerator())
             {
-                while (e1.MoveNext())
+                using (var e2 = other.Versions.GetEnumerator())
```
conniey: Biggest changes are that double nested using statements have to be separated.

PR #714, opened 2018-09-21
Last style cop rule fixes! Part 7

conniey:
Fixes:
* SA1612 - SA1614
* SA1616 - SA1620
* SA1621 - SA1628
* SA1629 - SA1632 
* SA1637 - SA1643
* SA1644 - SA1652

### DIFF
--- a/src/lib/Microsoft.Fx.Portability/IApiPortOptions.cs
+++ b/src/lib/Microsoft.Fx.Portability/IApiPortOptions.cs
@@ -12,6 +12,8 @@ namespace Microsoft.Fx.Portability
         string Description { get; }
 
         /// <summary>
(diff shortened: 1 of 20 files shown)


src/lib/Microsoft.Fx.Portability/IApiPortOptions.cs
```diff
@@ -12,6 +12,8 @@ public interface IApiPortOptions
         string Description { get; }
 
         /// <summary>
+        /// Gets a dictionary of input assembly files and whether or not they
+        /// were specified or not.
```
chlowell: "... and whether they were specified." ?

conniey: Fixed

PR #723, opened 2018-11-13
Update guidance for DefineDynamicAssembly

jkotas:


docs/RecommendedChanges/System/Use AssemblyBuilder.DefineDynamicAssembly.md
```diff
@@ -0,0 +1,5 @@
+### Recommended Action
+Use AssemblyBuilder.DefineDynamicAssembly
+
+### Affected APIs
+* `M:System.AppDomain.DefineDynamicAssembly`
```
terrajobst: @jkotas did you verify that this works? Doc IDs are usually matched by `==` and you probably to apply this to all overloads, right?

terrajobst: @conniey?

conniey: @terrajobst good catch. I'll create a PR to fix these.

PR #735, opened 2018-12-17
Fix file system test failure

twsouthwick:
Seems that not all tests are run on PR merges as this wasn't found until the dev CI pipeline was run.


twsouthwick:
Yeah... should look into it

PR #741, opened 2018-12-18
Merge dev into master

twsouthwick:


twsouthwick:
This will get a VSIX up that supports VS 2019

PR #742, opened 2018-12-18
Move HTML report generator to a razor library

### LINKED ISSUE #649: Render HTML reports with ASP.NET Core MVC
We can adapt the aspnet/Entropy [Mvc.RenderViewToString](https://github.com/aspnet/Entropy/tree/dev/samples/Mvc.RenderViewToString) sample to do this.

twsouthwick:
This change replaces RazorEngine with a ASP.NET Core Razor library. Since the MVC libraries only support .NET Standard 2.0, it will not run on .NET 4.6.1, so now HTML reports in offline mode are only supported on core.

Fixes #649


src/lib/Microsoft.Fx.Portability.Reports.Html/HtmlRazorReportWriter.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                options.FileProviders.Clear();
+                options.FileProviders.Add(fileProvider);
+            });
+
+            services.AddSingleton<ObjectPoolProvider, DefaultObjectPoolProvider>();
+            services.AddSingleton<DiagnosticSource>(new DiagnosticListener("Microsoft.AspNetCore"));
+            services.AddLogging();
+            services.AddMvc();
+            services.AddMvc();
```
chlowell: Redundant `AddMvc`

src/lib/Microsoft.Fx.Portability.Reports.Html/Views/_PortabilityReport.cshtml
```diff
@@ -51,7 +55,7 @@
         }
         @foreach (var invalidAssembly in reportingResult.GetAssembliesWithError())
         {
-            var fileName = Path.GetFileName(invalidAssembly);
+            var fileName = System.IO.Path.GetFileName(invalidAssembly);
```
conniey: Was there a collision when we had the `@using System.IO` in there?

twsouthwick: Yup - there's a property in razor apparently that is `Path`

src/lib/Microsoft.Fx.Portability.Reports.Html/HtmlRazorReportWriter.cs
twsouthwick: fixed

twsouthwick:
So the HTML writer is now available on .NET 4.6.1 builds as well as core - not sure why I was getting errors... probably a point in time thing with changing the projects around.

I also updated things to be non-async in #743 and merged those changes here to remove the pesky `.GetAwaiter().GetResut()`

PR #743, opened 2018-12-18
Change IReportWriter.WriteStream to be async

twsouthwick:


conniey:
I was just going to make a comment in your PR that it was too bad WriteReport wasn't async. Ahaha.

PR #748, opened 2019-01-23
Gather all our built binaries and symbols at one place

jeremymeng:
so we can run various tools on just binaries that we own

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -42,8 +42,6 @@
   <PropertyGroup>
     <IsTest>false</IsTest>
     <IsTest Condition="$(MSBuildProjectName.ToLowerInvariant().Contains('test'))">true</IsTest>
-    <IsVsix>false</IsVsix>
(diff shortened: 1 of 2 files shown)


jeremymeng:
Another option is removing the separate symbols folder and just place PDB files along side the assemblies.

Directory.Build.props
```diff
@@ -40,6 +42,8 @@
   <PropertyGroup>
     <IsTest>false</IsTest>
     <IsTest Condition="$(MSBuildProjectName.ToLowerInvariant().Contains('test'))">true</IsTest>
+    <IsVsix>false</IsVsix>
```
conniey: For ApiPort.Vsix, it sets a property that APiScan uses too, it is "<IsPackage>" on line 49.

jeremymeng: That's better.  Less changes. Fixed.

Issue #764, opened 2019-06-10
The HyperLinks to About and Privacy doc on Portability Analyzer Setting are not navigatable


Lxiamail:
In the VS extension, Portability Analyzer Setting dialog, the HyperLinks to "About Tool" and "View Privacy Statement" at the bottom of the dialog don't work.

Lxiamail:
Fixed by PR[#765](https://github.com/microsoft/dotnet-apiport/pull/765)

PR #765, opened 2019-06-10
The HyperLinks to About and Privacy docs on Portability Analyzer Setting are not navigatable

Lxiamail:
…Analyzer Setting are not navigatable

### DIFF
--- a/src/ApiPort/ApiPort.VisualStudio/Views/OptionsPageControl.xaml
+++ b/src/ApiPort/ApiPort.VisualStudio/Views/OptionsPageControl.xaml
@@ -186,7 +186,7 @@
 
             <TextBlock Grid.Column="2"
                        HorizontalAlignment="Right">
-                <Hyperlink RequestNavigate="NavigateToMoreInformation"
+                <Hyperlink NavigateUri="DocumentationLinks.About" RequestNavigate="NavigateToMoreInformation"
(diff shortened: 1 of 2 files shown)


src/ApiPort/ApiPort.VisualStudio/Views/OptionsPageControl.xaml
```diff
@@ -186,7 +186,7 @@
 
             <TextBlock Grid.Column="2"
                        HorizontalAlignment="Right">
-                <Hyperlink RequestNavigate="NavigateToMoreInformation"
+                <Hyperlink NavigateUri="DocumentationLinks.About" RequestNavigate="NavigateToMoreInformation"
```
twsouthwick: Is this needed? Looks like you either need to set the uri or could consolidate the event handlers

twsouthwick:
If you reference the issue in the body of the PR, it'll create a link. ie

#764

src/ApiPort/ApiPort.VisualStudio/Views/OptionsPageControl.xaml
Lxiamail: Yes, adding the uri actually fixed the original problem. Looks like consolidate the event handler only is not enough.

twsouthwick:
And if you say `Fixes ...` it'll automatically close the issue when merged

PR #779, opened 2019-07-03
fix MissingMemberInfo.UsedIn doesn't return anything

### LINKED ISSUE #776: MissingMemberInfo.UsedIn doesn't return anything
run offline mode "apiport analyze -f D:\GitRepos\dotnet-apiport\bin\Debug\ApiPort\net461\apiport.exe -o d:\temp\apiportresult -r Excel" got NullReferenceException from MissingMemberInfo.UsedIn. Looks like it was a regression introduced by was a bug introduced by this commit:https://github.com/microsoft/dotnet-apiport/commit/61e6aeec52cfbef08cb932745b2aa797f5cfcb62

Lxiamail:
Fix #776 . This also fixed the Usedby column in MissingAssembly worksheet is empty.

### DIFF
--- a/src/lib/Microsoft.Fx.Portability/Reporting/ObjectModel/MissingMemberInfo.cs
+++ b/src/lib/Microsoft.Fx.Portability/Reporting/ObjectModel/MissingMemberInfo.cs
@@ -12,7 +12,10 @@ namespace Microsoft.Fx.Portability.Reporting.ObjectModel
     {
         private readonly HashSet<AssemblyInfo> _usedInAssemblies;
 
-        public IEnumerable<AssemblyInfo> UsedIn { get; }
+        public IEnumerable<AssemblyInfo> UsedIn
+        {
+            get { return _usedInAssemblies; }
+        }
 
         public int Uses
         {
--- a/src/lib/Microsoft.Fx.Portability/Reporting/ObjectModel/MissingMemberInfo.cs
+++ b/src/lib/Microsoft.Fx.Portability/Reporting/ObjectModel/MissingMemberInfo.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Fx.Portability.Reporting.ObjectModel
 
         public IEnumerable<AssemblyInfo> UsedIn
         {
-            get { return _usedInAssemblies; }
+            get { return _usedInAssemblies.ToList().AsReadOnly(); }
         }
 
         public int Uses


src/lib/Microsoft.Fx.Portability/Reporting/ObjectModel/MissingMemberInfo.cs
```diff
@@ -12,7 +12,10 @@ public class MissingMemberInfo : MissingInfo
     {
         private readonly HashSet<AssemblyInfo> _usedInAssemblies;
 
-        public IEnumerable<AssemblyInfo> UsedIn { get; }
+        public IEnumerable<AssemblyInfo> UsedIn
+        {
+            get { return _usedInAssemblies; }
```
marklio: I'm not yet well-versed in the conventions of this codebase. I wouldn't hand _usedInAssemblies directly. Typically, the pattern for an accessor like this would be something like .ToList().AsReadOnly() to ensure the thing that gets handed out can't be modified. That's somewhat heavy-handed, but it's simple. Is there a sorting that would be nice to add here as well?
```suggestion
            get { return _usedInAssemblies.ToList().AsReadOnly(); }
```
Do you know if there are any multi-threading scenarios here that would lead to us being careful about accessing this while it's being modified? It doesn't look like that's a concern in this code.

Lxiamail: _usedInAssemblies is declared as readonly: "private readonly HashSet<AssemblyInfo> _usedInAssemblies;" The list is sorted at used time, I guess sorting at this place may be more optimized, however, I guess that I wouldn't change it now, and will change later if it appears needed more optimization. The code is used at excel report writing stage, doesn't look like there is any multi-threading scenarios to access this code.

twsouthwick: if you want to ensure someone can't just recast the return, since it's IEnumerable, you can just do an iterator:

```csharp
get
{
   foreach(var assembly in _usedInAssemblies)
   {
     yield return assembly; 
   }
```

This object I believe is used to built up a count of how many times something is used and is run in a Parallel.ForEach

twsouthwick:
Your title doesn't match the issue...

src/lib/Microsoft.Fx.Portability/Reporting/ObjectModel/MissingMemberInfo.cs
Lxiamail: @twsouthwick that's good suggestion. Since the risk of recast the return is low, I would rather to keep the code simple as it is for now if it looks good to you.
@Marklio Are you good with the current code to merge?

marklio: It's fine with me.  Again, my suggestions are based on principles I use in my own code to stay out of trouble WRT thread safety and immutability. We'll have to decide what principles we want to apply to this codebase as time goes one. For the sake of discussion, @twsouthwick, iterator is a reasonable approach to prevent exposing the underlying storage, but it will throw if the storage is updated during an iteration, while a readonly snapshot won't.

Lxiamail: Understood @marklio's concern. committed his suggested change. Good discussion!

PR #783, opened 2019-07-09
Consistent label and narration for report file name field

chlowell:
The label and automation name for the default report file name field of the settings control don't match, causing screen readers to describe the field inconsistently with its label. This makes both the label and the automation name "Default report file name".


twsouthwick:
Haha - I'm adding an option and was wondering about that :)

PR #785, opened 2019-07-10
Cache Version and FrameworkName in json deserialization

twsouthwick:
This converter is used to create Version and FrameworkName instances which are a small set of known values in the assembly. This caches them so that we can reuse them when creating them.

### DIFF
--- a/src/lib/Microsoft.Fx.Portability/Utils/JsonConverters/JsonToStringConverter.cs
+++ b/src/lib/Microsoft.Fx.Portability/Utils/JsonConverters/JsonToStringConverter.cs
@@ -4,15 +4,18 @@
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using System;
+using System.Collections.Concurrent;
 
 namespace Microsoft.Fx.Portability.Utils.JsonConverters
 {
     internal class JsonToStringConverter<T> : JsonConverter<T>
     {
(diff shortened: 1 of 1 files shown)


src/lib/Microsoft.Fx.Portability/Utils/JsonConverters/JsonToStringConverter.cs
```diff
@@ -4,23 +4,26 @@
...
+            _cache = new ConcurrentDictionary<string, T>(StringComparer.Ordinal);
             _factory = factory;
         }
 
         public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
         {
             var value = serializer.Deserialize<JToken>(reader).ToString();
 
-            return string.IsNullOrEmpty(value) ? default(T) : _factory(value);
+            return string.IsNullOrEmpty(value) ? default : _cache.GetOrAdd(value, _factory);
```
Lxiamail: we still want "default(T)" instead of "default", right?

conniey: `default` is the literal keyword in C# 7.1 for `default(T)`
https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7-1#default-literal-expressions

PR #788, opened 2019-07-13
fix binskim bug for HighEntropyVACheck issue

Lxiamail:
bug title: [ VS-DotNet-ApiPort ][ BinSkim ] - Defect : HighEntropyVACheck, Component : \a\2.6.0-alpha.10\ApiPort\netcoreapp2.1\ApiPort.dll (1 issue)
Same issue for offline version as well.
Binskim error:
error BA2015: 'ApiPort.dll' does not declare itself as high entropy ASLR compatible. High entropy makes Address Space Layout Randomization more effective in mitigating memory corruption vulnerabilities. To resolve this issue, configure your tools to mark the program high entropy compatible; e.g. by supplying /HIGHENTROPYVA to the C or C++ linker command line. (This image was determined to have been properly compiled as /LARGEADDRESSAWARE.)

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -64,13 +64,14 @@
     </PackageReference>
   </ItemGroup>
 
-  <!-- Work around an issue where portable PDB is not supported yet in Api scan tool.
+  <PropertyGroup>
+<!-- Work around an issue where portable PDB is not supported yet in Api scan tool.
        This does not work in Directory.Build.props because $(TargetFramework) has not been set yet
        when it is imported. -->
-  <PropertyGroup>
     <DebugType Condition=" $(TargetFramework.StartsWith('net4')) ">Full</DebugType>
+    <HighEntropyVA>true</HighEntropyVA>
   </PropertyGroup>
-
+   
   <!-- In .NET Core SDK 1.0-rc3, all *.cs files are implicitly added for compiled.  This includes the test files we embed as resources.
        We explicitly remove them here so that they are not compiled.
        BUG: https://github.com/dotnet/sdk/issues/977 -->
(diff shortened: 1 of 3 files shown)


Lxiamail:
@marklio @twsouthwick bubble up the CR request.

marklio (commented):
Is this actually necessary? My understanding is that this is on by default.

Lxiamail:
I was wandering if it was enabled by default. However, the binskim bug auto opened by the build system. I also searched some other repos and found some with this setting specifically enabled.

twsouthwick:
Never had to use this, so not sure what the default is. But you'd want to put this in the `Directory.Build.props` so it's picked up by all projects

Lxiamail:
> Never had to use this, so not sure what the default is. But you'd want to put this in the `Directory.Build.props` so it's picked up by all projects
Good point. Fixed

marklio:
I know csc doesn't do it by default, but the targets should be setting it if you are targeting a supported framework: https://github.com/microsoft/msbuild/blob/a78f69dae3d599544e52b69b15e5bd71e27cd3b3/src/Tasks/Microsoft.NETFramework.CurrentVersion.props#L108

marklio:
I'm looking into why this behavior isn't working.

marklio:
Looks like this tried to get fixed with https://github.com/microsoft/msbuild/issues/2912, but the fix wasn't correct. It is now being tracked by https://github.com/microsoft/msbuild/issues/4566

marklio (approved):

twsouthwick (approved):

terrajobst:
Closing as API Port was deprecated in favor of binary analysis in [.NET Upgrade Assistant](https://github.com/dotnet/upgrade-assistant/blob/main/docs/binary_analysis.md).

Issue #668, opened 2018-05-31
Publish the client as a .NET Core global tool


jeremymeng:
https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools

It makes it easier to consume for developers.

conniey:
Related to #240

terrajobst:
@Lxiamail any way we could prioritize this?

Lxiamail:
@terrajobst I planned to work on some report related improvement work items first since we have heard some customer feedback. Have we hears customer requests for this issue? If so, we can certainly prioritize this.

terrajobst:
The benefit of doing it this way is that customers can get updates more quickly. Many people are still confused on how to acquire it. I keep getting pinged on why it's not a Global Tool yet.

@KathleenDollard @danroth27

Lxiamail:
@terrajobst This is definitely something that we should do soon.

KathleenDollard:
If it is a .NET Core shared framework dll, then creating a global tool is a fairly painless process, although creating a signed build (required) may not be. 

If it is a shared framework dll already, I can help with the syntax of the CLI and naming, and with the System.CommandLine parser. With this, I think the work to create the tool will be smallish. 

If I tis not a shared framework dll, that would be a problem.

Lxiamail:
@KathleenDollard It is not shared framework dll. It is only released in a zip file from this repo. That is the pain point that we want to address.

Lxiamail:
@KathleenDollard is https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools-how-to-create the document that I should follow to create a global tool?

KathleenDollard:
That is a good start. I am working on another document for MSFT authored tools. Here are additional comments.


  *   All .NET (Global) Tools are framework dependent executables
     *   We are working on guidelines for 2.1/3.0. Suggest testing on both and Muti-targeting
  *   Highly recommend you create your help in word or markdown first
     *   It should parallel dotnet --help
     *   I highly recommend you review this syntax with me for consistency
  *   Highly recommend you use System.CommandLine for parsing. See https://github.com/dotnet/command-line-api
     *   Jon Sequeira can answer questions
  *   Signing is required. William Li may be able to help some
  *   The package name must be prefixed with "Microsoft". You must also reserve the non-Microsoft prefixed version.

I'm happy to schedule some time with you on this. I am out Friday and next week. I come in early if you are an early person.
________________________________

[...]

PR #790, opened 2019-07-22
Make ApiPort to be Global tool. More version updates

Lxiamail:
This PR includes the following changes:
Make ApiPort to be Global tool (issue https://github.com/microsoft/dotnet-apiport/issues/668).
Updated ApiPort console apps to .NET Core 2.1 standalone app and removed desktop app support;
Updated ApiPort libraries to remove NET standard1.3 support and only support .NET standard2.0; 
Update VS extensions to target .net4.6.1 to support .NET standard. 
Fixed some warnings to call some APIs with specific cultureInfo.


src/ApiPort/ApiPort/ApiPort.props
```diff
@@ -3,20 +3,29 @@
...
+  <Choose>
+    <!-- build as global tool if GlobalTool is specified-->
+    <When Condition="'$(GlobalTool)' == 'true'">
+      <PropertyGroup>
+      <PackAsTool>true</PackAsTool>
+      <ToolCommandName>ApiPort</ToolCommandName>
+     <!-- <PackageOutputPath>./nupkg</PackageOutputPath> -->
+        </PropertyGroup>
+    </When>
+    <!-- build as standardalone .NET Core app by default-->
```
KathleenDollard: nit: standalone

src/ApiPort/ApiPort.VisualStudio/ApiPort.VisualStudio.csproj
```diff
@@ -10,7 +10,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>ApiPortVS</RootNamespace>
     <AssemblyName>ApiPort.VisualStudio</AssemblyName>
-    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
```
twsouthwick: i think VS has a minimum of 4.6, so we don't want to rely on something later unless we have to.

src/ApiPort/ApiPort.VisualStudio/ApiPort.VisualStudio.csproj
```diff
@@ -137,7 +137,9 @@
     </ProjectReference>
     <ProjectReference Include="..\..\lib\Microsoft.Fx.Portability\Microsoft.Fx.Portability.csproj">
       <Project>{8d84ec23-9977-4cc8-b649-035ffae9664c}</Project>
+      <!--Looks like net64 specified was because Microsoft.Fx.Portability TF was net46. Was there any reason it can't be net461 or newer?
```
twsouthwick: Without this, VS extensions don't correctly bundle multi-targeting apps

src/ApiPort/ApiPort.Vsix/ApiPort.Vsix.csproj
```diff
@@ -11,7 +11,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>ApiPort.Vsix</RootNamespace>
     <AssemblyName>ApiPort.Vsix</AssemblyName>
-    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
```
twsouthwick: Again, verify what the minimum support for VS 2017 is

src/ApiPort/ApiPort/ApiPort.props
```diff
@@ -3,20 +3,29 @@
...
     <AssemblyName>ApiPort</AssemblyName>
     <RootNamespace>ApiPort</RootNamespace>
-    <!--
-      netcoreapp target must be first to avoid a ResXFileCodeGenerator issue
-      (tracked at https://github.com/dotnet/project-system/issues/1519)
-    -->
-    <TargetFrameworks>netcoreapp2.1;net461</TargetFrameworks>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
     <GenerateBindingRedirectsOutputType>true</GenerateBindingRedirectsOutputType>
```
twsouthwick: Verify proxy support now that you're removing this. I believe this is the only thing that would not be supported (@conniey can you think of anything else)?

src/ApiPort/ApiPort/ApiPort.props
```diff
@@ -3,20 +3,29 @@
...
+      <PackAsTool>true</PackAsTool>
+      <ToolCommandName>ApiPort</ToolCommandName>
+     <!-- <PackageOutputPath>./nupkg</PackageOutputPath> -->
+        </PropertyGroup>
+    </When>
+    <!-- build as standardalone .NET Core app by default-->
+    <Otherwise>
+      <PropertyGroup>
+        <RuntimeIdentifiers>win7-x64;win7-x86</RuntimeIdentifiers>
+        <!--TODO: These constants was defined when building as 461 app, looks out of date. Do we still need to define these constants?
```
twsouthwick: These aren't needed now AFAIK - it's auto added I believe

src/lib/Microsoft.Fx.Portability.Cci/TypeExtensions.cs
```diff
@@ -609,7 +609,7 @@ public static bool IsReferenceAssembly(this IAssembly assembly)
 
         public static string MapPublicKeyTokenToName(string publicKeyToken)
         {
-            if (!TokenNames.TryGetValue(publicKeyToken.ToLower(), out var name))
+            if (!TokenNames.TryGetValue(publicKeyToken.ToLowerInvariant(), out var name))
```
twsouthwick: Not a blocker for this, but the `TokenNames` field should probably be created with a `StringComparer.OrdinalIgnoreCase` comparer so no string manipulations are needed

tests/ApiPort/ApiPort.Tests/ApiPort.Tests.csproj
```diff
@@ -1,7 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <!--
```
twsouthwick: why not remove this?

src/ApiPort/ApiPort.VisualStudio/ApiPort.VisualStudio.csproj
Lxiamail: Per this document https://docs.microsoft.com/en-us/visualstudio/productinfo/vs2017-system-requirements-vs, .NET 4.6.1 is installed with VS 2017 by default. Looks like we should be good. The reason to update to 4.6.1 is to meet the minimum .NET version requirement to support .NET Standard 2.0.

src/ApiPort/ApiPort/ApiPort.props
Lxiamail: Looks like FEATURE_SYSTEM_PROXY and FEATURE_NETWORK_CREDENTIAL was defined when targeting 4.6.1 was because  WebRequest.GetSystemWebProxy() and NetworkCredential.SecurePassword were not available on .NET Standard 1.3. They are supported on .NET standard 2.0+. Since this PR removed .NET standard 1.3 support,  looks like it is okay to remove both defineconstants.

src/lib/Microsoft.Fx.Portability.Cci/TypeExtensions.cs
Lxiamail: good point! Fixed

src/ApiPort/ApiPort/ApiPort.props
Lxiamail: @twsouthwick Actually. I said too fast, SystemWebProxy.GetProxy throws platformnotsupported exception. This means that the code under ifdef FEATURE_SYSTEM_PROXY still only work on desktop. With this PR, we actually removing desktop support. What would be the customer visible missing functionality on .NET Core then?

Lxiamail: @twsouthwick added net461 targetframework back to the console tools when building self-contained because FEATURE_SYSTEM_PROXY still only works on desktop, we don't want to regress existing functionality

src/ApiPort/ApiPort.Vsix/ApiPort.Vsix.csproj
Lxiamail: minimum support is 4.6.1 as stated above.

src/ApiPort/ApiPort/ApiPort.props
Lxiamail: Added back desktop support to make sure not regress proxy feature.

Lxiamail:
@twsouthwick Do you have any additional feedback?

src/lib/Microsoft.Fx.Portability.Cci/TypeExtensions.cs
marklio: For the record, I don't see this fix in the PR.

src/lib/Microsoft.Fx.Portability.MetadataReader/MemberDependency.cs
```diff
@@ -106,7 +106,7 @@ public override int GetHashCode()
         {
             if (!_hashComputed)
             {
-                _hashCode = ((DefinedInAssemblyIdentity?.ToString() ?? string.Empty) + (MemberDocId ?? string.Empty) + IsPrimitive.ToString()).GetHashCode() ^ CallingAssembly.GetHashCode();
+                _hashCode = ((DefinedInAssemblyIdentity?.ToString() ?? string.Empty) + (MemberDocId ?? string.Empty) + IsPrimitive.ToString((IFormatProvider)CultureInfo.CurrentUICulture)).GetHashCode() ^ CallingAssembly.GetHashCode();
```
marklio: Not a blocker, but I don't think CurrentUICulture is appropriate for a hashcode. We should use Invariant here. As a point of preference, this seems like a somewhat iffy way to construct a hashcode. I think we have some hashcode helpers now. I'll see if I can find them.

marklio: Ah. Only in Core/ns2.1: https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=netstandard-2.1

marklio: Use InvariantCulture

Lxiamail: fixed

Lxiamail:
Test failures due to missing ilasm is an known SDK issue and fixed in SDK 3.0. more details at https://github.com/dotnet/cli/issues/12341. we can ignore the failures for now.

PR #792, opened 2019-07-31
Add titles and labels to HTML tables

chlowell:
This gives each table in the HTML report a title or aria label, so screen readers have a name to announce them by.


src/lib/Microsoft.Fx.Portability.Reports.Html/Views/_BreakingChangesReport.cshtml
```diff
@@ -18,8 +18,8 @@
 
 <h2 id="@LocalizedStrings.CompatibilityPageTitle" tabindex="0">@LocalizedStrings.CompatibilityPageTitle</h2>
 <div id="summary-compat-byassembly">
-    <h3 class="compat-subheader">@LocalizedStrings.PerAssemblyOverview</h3>
-    <table>
+    <h3 class="compat-subheader" id="compat-byassembly-subheader">@LocalizedStrings.PerAssemblyOverview</h3>
```
Lxiamail: With this change, will @LocalizedStrings.PerAssemblyOverview be read two times, one as h3 and one as table. Also per https://www.accessibility-developer-guide.com/examples/sensible-aria-usage/label-labelledby/, aria-label and aria-labelledby has some undesirable side effects. What if remove <h3 class="compat-subheader">, and add caption with @LocalizedStrings.PerAssemblyOverview as value to the table?

src/lib/Microsoft.Fx.Portability.Reports.Html/Views/_BreakingChangesReport.cshtml
```diff
@@ -81,8 +81,8 @@
 @if (Model.OrderedBreakingChangeSkippedAssemblies.Count() > 0)
 {
     <div id="summary-compat-skippedassemblies">
-        <h3 class="compat-subheader">@LocalizedStrings.IgnoredAssemblies</h3>
-        <table>
+        <h3 class="compat-subheader" id="compat-skippedassemblies-subheader">@LocalizedStrings.IgnoredAssemblies</h3>
```
Lxiamail: Save feedback as above

src/lib/Microsoft.Fx.Portability.Reports.Html/Views/_PortabilityReport.cshtml
```diff
@@ -22,7 +22,7 @@
 </svg>
 
 <h2 id="@LocalizedStrings.PortabilitySummaryPageTitle" tabindex="0">@LocalizedStrings.PortabilitySummaryPageTitle</h2>
-<table>
+<table aria-labelledby="@LocalizedStrings.PortabilitySummaryPageTitle">
```
Lxiamail: same comment as above

src/lib/Microsoft.Fx.Portability.Reports.Html/Views/_PortabilityReport.cshtml
```diff
@@ -114,7 +114,7 @@
         var usedUnresolvedAssembly = Model.GetUnresolvedAssemblies(assembly);
         if (usedUnresolvedAssembly.Count > 0)
         {
-            <table>
+            <table aria-labelledby="@assembly.SourceAssembly">
```
Lxiamail: Looks like this table is for missing assemblies. What if instead of adding aria-labelledby="@assembly.SourceAssembly", adding a caption with value of @LocalizedStrings.MissingAssemblies? Certainly, that could seems duplicated with the table header  of LocalizedStrings.MissingAssemblies .

src/lib/Microsoft.Fx.Portability.Reports.Html/Views/_BreakingChangesReport.cshtml
chlowell: `<caption>` works but Narrator reads it twice in Edge, once in Firefox. I agree it's preferable semantically, so I'll push a commit labeling the tables with `<caption>`.

Issue #795, opened 2019-08-14
DGML fails when there is nonresolved dependency 


Lxiamail:

Lxiamail:
fixed

PR #796, opened 2019-08-14
DGML fails when there is unresolved dependency

Lxiamail:
Fix issue #795

### DIFF
--- a/src/lib/Microsoft.Fx.Portability.Reports.DGML/ReferenceNode.cs
+++ b/src/lib/Microsoft.Fx.Portability.Reports.DGML/ReferenceNode.cs
@@ -105,6 +105,12 @@ namespace Microsoft.Fx.Portability.Reports.DGML
 
             foreach (var item in Nodes)
             {
+                if (item.UsageData == null)
+                {
+                    // skip the Node with no UsageData
+                    continue;
+                }
+
                 if (!item.TryGetAPICountFromReferences(target, out int refCountAvail, out int refCountUnavail))
                 {
                     // Cycle!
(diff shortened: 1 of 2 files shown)


src/lib/Microsoft.Fx.Portability.Reports.DGML/ReferenceNode.cs
```diff
@@ -105,6 +105,12 @@ public bool TryGetAPICountFromReferences(int target, out int availAPIs, out int
 
             foreach (var item in Nodes)
             {
+                if (item.UsageData == null)
+                {
+                    // skip the Node with no UsageData
+                    continue;
```
marklio: This will potentially no longer return false in cycle cases. That seems like maybe it's OK because all such cases would crash due to null ref, but we would no longer do the cycle checking. I would  move this check to the "leaf" call (just before setting availAPIs and unavailAPIs), and still do the _searchInGraph check. Depending on how TryGetAPICountFromReferences works, there may be other bugs in this recursive method (we will stop counting sibling nodes as soon as we detect a cycle, which doesn't seem right to me).

Lxiamail: what your comment makes sense. Looks like there is another bug in the original below code the return false when cycle is detected may stop counting the api counts of remaining nodes in "Nodes" collection.  
`
                if (!item.TryGetAPICountFromReferences(target, out int refCountAvail, out int refCountUnavail))
                {
                    // Cycle!
                    _searchInGraph = false; // Reset this flag
                    return false;
       }
`
Since this is a separated existing issue, I will treat as separated issue without blocking the current checkin. When we address this issue, I may need to reconsider where the usagedata null check should be.

marklio: Just so we remember, if an item can have null UsageData and non-Empty Nodes, we'll want to enumerate the nodes even if we skip UsageData.

Issue #799, opened 2019-09-07
PublicKeyToken.Token  doesn't serialized to server side correctly


Lxiamail:
PublicKeyToken.Token is always null after deserialized on the server side.

PR #800, opened 2019-09-07
Fix public key token serialization issue

Lxiamail:
#799, which is a breaking change introduced by PR #784

### DIFF
--- a/src/lib/Microsoft.Fx.Portability/Utils/JsonConverters/JsonPublicKeyTokenConverter.cs
+++ b/src/lib/Microsoft.Fx.Portability/Utils/JsonConverters/JsonPublicKeyTokenConverter.cs
@@ -9,18 +9,18 @@ using System.Text;
 namespace Microsoft.Fx.Portability.Utils.JsonConverters
 {
     internal class JsonPublicKeyTokenConverter : JsonConverter<PublicKeyToken>
-    { 
+    {
         public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
         {
-            ImmutableArray<byte> bytes = serializer
-                .Deserialize<ImmutableArray<byte>>(reader);
-            return new PublicKeyToken(bytes);
+            string publicKeyTokenString = serializer
+                .Deserialize<string>(reader);
+            return PublicKeyToken.Parse(publicKeyTokenString);
         }
 
         public override void WriteJson(JsonWriter writer, object obj, JsonSerializer serializer)
         {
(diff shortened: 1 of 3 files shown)


src/lib/Microsoft.Fx.Portability/Utils/JsonConverters/JsonPublicKeyTokenConverter.cs
```diff
@@ -0,0 +1,26 @@
...
+using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+
+namespace Microsoft.Fx.Portability.Utils.JsonConverters
+{
+    internal class JsonPublicKeyTokenConverter : JsonConverter<PublicKeyToken>
```
marklio: Where is this serialization used and what is it optimized for? Serializing as a string would be a much more efficient and easy to deal with encoding, but this looks like it would work.

twsouthwick: It's probably used in returning a response or sending a request. Must not have any tests associated with it so my change wasn't caught.

Lxiamail: @twsouthwick This PR fix the current problem if both client and server sides using the new binary contains PublicKeyToken struct, but if old version of client without knowledge of PublicKeyToken works with new Server, the serialization will still break. Looks like PR #784 broke backward compat serialization.

twsouthwick: You should be able to have a converter that would work for both. Do a string serialization (maybe that's what @marklio was referring to) and it should then work for backwards compatibility

twsouthwick: You can serialize with `PublicKeyToken.ToString()` and deserialize with `PublicKeyToken.Parse(string)`. That should provide round tripping

Lxiamail: Submitted a new commit. Verified working with backward compat.

Lxiamail:
dotnet-appport-dev-CI failed 14 tests for known SDK issue of ilasm is missing.

PR #805, opened 2019-10-10
Created two tools for updating recommended change markdown files

Lxiamail:
RecommendedChangeMarkDownReader: Read recommended change markdown files to an excel file for each editing
RecommendedChangeMDWriter: Create markdown files from an excel file. The markdown files are organized by namespaces and unique recommended changes.


marklio:
Any reason you're not just making a new repo for these?

Lxiamail:
> Any reason you're not just making a new repo for these?

It may be a better option. I thought of checking into this repo is because the md files are in the same repo. However, they don't have to be in the same repo, separating these tools from product repo may be better choose.

PR #806, opened 2019-10-16
Update global.json file

Lxiamail:
to specify sdk version 2.2 to avoid build error when build with sdk 3.0


Lxiamail:
Since the build machine may have the specified SDK version installed, we shouldn't checkin the global.son file with the specified SDK version.

PR #808, opened 2019-10-18
Add reference to Microsoft.NETCore.Platforms package to workaround A SDK issue

Lxiamail:
There is a SDK issue in .NET Core 2.* (https://github.com/dotnet/cli/issues/12341). The issue results with ilasm.exe is not copied to Microsoft.Fx.Portability.MetadataReader.Tests\net461\win7-x86 output directory as it should be. Add reference to Microsoft.NETCore.Platforms package to workaround the issue. The issue has been fixed in .NET Core 3.0 SDK (verified on 3.0.100). The reference should be removed once build machine moves to .NET Core 3.0.


Lxiamail:
The test failure is not related with this PR and has been fixed.

PR #807, opened 2019-10-16
Create nuget package at build time

Lxiamail:
to make it easy for portabilityservice to consume at local development time.

PR #751, opened 2019-01-24
Reports.Html works on .NET Framework

conniey:
In .NET Framework, the precompiled views are not added to the ApplicationParts by default, so they cannot be resolved.

Fixes #749

PR #746, opened 2018-12-20
Adding support for running tests cross-platform

conniey:
* This is so we don't have to exclude random tests when running our CI/CD builds in our build definition
* Adds FEATURE_ILASM for running tests on .NET 4.6.1 (it should run on .NETCoreApp2.1) but for some reason, native assets aren't copied over when running on .NETCoreApp2.1. https://github.com/dotnet/sdk/issues/2765

PR #745, opened 2018-12-18
Add DiagnosticListener to report ServiceProvider

conniey:
Fixes #744 

This came up because our CI pipeline was only running a subset of the tests. Should be fixed now.

PR #809, opened 2019-10-22
Integrate changes from dev branch to master

Lxiamail:
**Functional improvements:**
Fixed all Y11MAS accessibility issues to meet grade C requirements (#783, #792)
Make ApiPort tool to be .NET Core Global tool (#790)
Fixed DGML fails when there is unresolved dependency(#796)
Moved HTML report generator to a razor library (#742)
Added DiagnosticListener to report ServiceProvider (#745)
Update TargetNameParser to understand versions (#757)
Reports.Html works on .NET Framework (#751)
Updated the HyperLinks to About and Privacy docs on Portability Analyzer Setting are not navigable (#765)
Updated the document link in excel report (#807)
Fix Assembly name header name in Excel report (#779)
Fixed MissingMemberInfo.UsedIn doesn't return anything (#779)
Added JsonPublicKeyTokenconverter to Serialize PublicKeyToken (#800)

**Fundamental improvements**:
Change IReportWriter.WriteStream to be async (#743)
Don't require AssemblyReferenceInformation for IDependencyFilter to reduce memory allocation (#784)
Upgraded .NET Portability libraries from .NET Standard 1.3 to .NET Standard 2.0 (#790)
Upgraded ApiPort Console apps to .NET Core 2.1 and .NET Framework 4.6.1. Upgraded ApiPort VS extension to 4.6.1 (#790)
Set up some caching in json conversion (#785)

**Engineering improvements**:
Update StyleCop.Analyzers and compiler version used (#747)
Gather all our built binaries and symbols at one place (#748) 
Adding support for running tests cross-platform (#746)
Fiximg LicenseUrl is deprecated warning (#750)
Use Full PDB format for net4* projects (#754)
Add reference to Microsoft.NETCore.Platforms to workaround dotnet/cli issue… (#808)


NuGet.config
```diff
@@ -4,8 +4,6 @@
     <!-- Don't use any sources set by the system (workaround for microbuild issue) -->
     <clear />
     <add key="nuget" value="https://www.nuget.org/api/v2/" />
-    <add key="myget.org dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
     <add key="corefx lab" value="https://dotnet.myget.org/F/dotnet-corefxlab/api/v3/index.json" />
```
marklio: Do we still need this source?

src/ApiPort/ApiPort.VisualStudio.Common/ComProjectMapper.cs
```diff
@@ -16,7 +16,7 @@
 namespace ApiPortVS
 {
     /// <summary>
-    /// You need to switch to the UI thread before calling any COM interfaces (e.g. IVsHierarchy) using
+    /// You need to switch to the UI thread before calling any COM interfaces (e.g. IVsHierarchy) using.
```
marklio: It feels like maybe there used to be another line here so that it said something like "...using ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync()".  As is, it doesn't make sense, and the period doesn't help.

src/ApiPort/ApiPort.VisualStudio/AssemblyRedirectResolver.cs
```diff
@@ -16,7 +16,7 @@ namespace ApiPortVS
     /// Because the VSIX runs inside devenv.exe, the only configuration file that is used for redirects is
     /// devenv.exe.config.We don't want to modify this machine wide configuration so we are manually setting
     /// the redirects.
-    /// More info: http://stackoverflow.com/a/31248093/4220757
+    /// More info: http://stackoverflow.com/a/31248093/4220757.
```
marklio: I'm not a fan of these rules that make you put periods everywhere. This isn't a sentence, so it doesn't need a period. In this case, depending on the interpretation/viewer of the URL, it might include the period and end up with a broken URL. It looks like stackoverflow handles this fine, so I'll leave it to you, but I think these are low-value.

src/lib/Microsoft.Fx.Portability.Cci/Analyzer/DependencyFinderEngine.cs
```diff
@@ -60,22 +60,30 @@ private void FindDependencies(IProgressTask progressTask)
...
+                            MemberDocId = dep.MemberDocId,
+                            TypeDocId = dep.TypeDocId,
+                            DefinedInAssemblyIdentity = dep.DefinedInAssemblyIdentity
+                        };
 
                         // Add this memberinfo
                         var newassembly = new HashSet<AssemblyInfo>
                         {
                             dep.CallingAssembly
                         };
```
marklio: Consider using the overload of AddOrUpdate that takes a factory for the initial value as well, to avoid allocating the HashSet whether you use it or not.

src/lib/Microsoft.Fx.Portability.Cci/Analyzer/DependencyFinderEngine.cs
```diff
@@ -60,22 +60,30 @@ private void FindDependencies(IProgressTask progressTask)
...
-                                                                                                            lock (existingSet)
-                                                                                                            {
-                                                                                                                existingSet.Add(dep.CallingAssembly);
-                                                                                                            }
-                                                                                                            return existingSet;
-                                                                                                        });
+
+                        var assemblies = dependencies.AddOrUpdate(m, newassembly, (key, existingSet) =>
+                        {
+                            lock (existingSet)
```
marklio: Consider removing this lock. ConcurrentDictionary ensures that updates to the values done in the factories are "atomic".

src/lib/Microsoft.Fx.Portability.Cci/Analyzer/DependencyFinderEngine.cs
```diff
@@ -99,16 +107,15 @@ private IEnumerable<MemberDependency> GetDependencies(string assemblyLocation)
...
 
-                    HashSet<string> newValue = new HashSet<string>
-                    {
-                        callingAssembly
-                    };
+                    var newValue = new HashSet<string> { callingAssembly };
+
                     _unresolvedAssemblies.AddOrUpdate(e.Unresolved.Format(), newValue, (key, existingHashSet) =>
                     {
                         lock (existingHashSet)
```
marklio: More unneeded locks. I guess maybe don't mess with this too much. We can make another pass at some point to simplify.

src/lib/Microsoft.Fx.Portability.MetadataReader/MemberDependency.cs
```diff
@@ -106,7 +106,7 @@ public override int GetHashCode()
         {
             if (!_hashComputed)
             {
-                _hashCode = ((DefinedInAssemblyIdentity?.ToString() ?? string.Empty) + (MemberDocId ?? string.Empty) + IsPrimitive.ToString()).GetHashCode() ^ CallingAssembly.GetHashCode();
+                _hashCode = ((DefinedInAssemblyIdentity?.ToString() ?? string.Empty) + (MemberDocId ?? string.Empty) + IsPrimitive.ToString((IFormatProvider)CultureInfo.InvariantCulture)).GetHashCode() ^ CallingAssembly.GetHashCode();
```
marklio: Why are we ToString-ing something that appears to bee a bool?

src/lib/Microsoft.Fx.Portability.MetadataReader/MemberMetadataInfo.cs
```diff
@@ -239,7 +239,7 @@ private IEnumerable<string> GetGenericDisplayNames(IList<string> displayNames)
         }
 
         /// <summary>
-        /// Replace the non-safe characters (<>.,) in a method name with safe ones ({}#@)
+        /// Replace the non-safe characters (.<>.,) in a method name with safe ones ({}#@)
```
marklio: It looks like period is already in the list? Was this another application of the period rule?

NuGet.config
Lxiamail: Yes, we still need it. Apparently,  we never released the microsoft.fx.portablity.* nuget packages on nuget.org, They are only available on myget.

src/ApiPort/ApiPort.VisualStudio/AssemblyRedirectResolver.cs
Lxiamail: Agreed that it is annoying and low value. Added all the periods due to azure devops build machine reporting this style warning as errors. I don't see which rule forces the ending period from https://github.com/microsoft/dotnet-apiport/blob/dev/stylecop.json. Any idea where it may come from?

marklio: Those are driven by the new style analyzers written in Roslyn (rather than stylecop). You can configure them with .editorconfig. You should be getting the same messages in your local editor/build. There should be a way to configure them to not be noisy by right-clicking them and suppressing them. I think that this project is opting into them (I don't think you get them by default)

src/lib/Microsoft.Fx.Portability.MetadataReader/MemberDependency.cs
Lxiamail: Not sure why. The only thing that I can think of may be hashcodes of string "true" vs "false" is more specific than bool values (less chance of ambiguity).

src/lib/Microsoft.Fx.Portability.Cci/Analyzer/DependencyFinderEngine.cs
Lxiamail: I would prefer to take care of these in another pass.  Open https://github.com/microsoft/dotnet-apiport/issues/810 to track the issue.

src/lib/Microsoft.Fx.Portability.Cci/Analyzer/DependencyFinderEngine.cs
Lxiamail: I would prefer to take care of these in another pass.  Open https://github.com/microsoft/dotnet-apiport/issues/810 to track the issue.

src/lib/Microsoft.Fx.Portability.Cci/Analyzer/DependencyFinderEngine.cs
Lxiamail: I would prefer to take care of these in another pass.  Open https://github.com/microsoft/dotnet-apiport/issues/810 to track the issue.

src/ApiPort/ApiPort.VisualStudio.Common/ComProjectMapper.cs
Lxiamail: Fixed by PR https://github.com/microsoft/dotnet-apiport/pull/811

src/ApiPort/ApiPort.VisualStudio/AssemblyRedirectResolver.cs
Lxiamail: Fixed by PR https://github.com/microsoft/dotnet-apiport/pull/811

src/lib/Microsoft.Fx.Portability.Reports.DGML/DGMLOutputWriter.cs
```diff
@@ -26,7 +24,7 @@ public class DGMLOutputWriter : IReportWriter
 
         private readonly DGMLManager dgml = new DGMLManager();
 
-        public void WriteStream(Stream stream, AnalyzeResponse response)
+        public Task WriteStreamAsync(Stream stream, AnalyzeResponse response)
```
marklio: Nice To Have: Useless change to async. This method always completes synchronously.

src/lib/Microsoft.Fx.Portability.Reports.DGML/ReferenceNode.cs
```diff
@@ -105,6 +105,12 @@ public bool TryGetAPICountFromReferences(int target, out int availAPIs, out int
 
             foreach (var item in Nodes)
             {
+                if (item.UsageData == null)
```
marklio: I feel like I've reviewed this or similar changes before. I'm concerned that this check and continue will prevent recursion (and therefore proper counting) if a Node without usage data, but containing Nodes is encountered. Ideally, the UsageData check should be around the initial calculation of avail/unavailAPIs, but still allow recursion. I'll see if I can propose a change.

marklio: At the top...
```suggestion
                if (UsageData is null)
                {
                    availAPIs = 0;
                    unavailAPIs = 0
                }
                else
                {
                    availAPIs = UsageData[target].GetAvailableAPICalls();
                    unavailAPIs = UsageData[target].GetUnavailableAPICalls();
                }
```

marklio: Bah, silly github. That's not the diff I want. This block should go away, and the top of the method should look like the green above.

src/lib/Microsoft.Fx.Portability.Reports.Json/JsonReportWriter.cs
```diff
@@ -5,29 +5,25 @@
...
         }
 
-        public ResultFormatInformation Format
-        {
-            get { return _formatInformation; }
-        }
+        public ResultFormatInformation Format { get; }
 
-        public void WriteStream(Stream stream, AnalyzeResponse response)
+        public Task WriteStreamAsync(Stream stream, AnalyzeResponse response)
```
marklio: Nice to have: useless async change.

src/lib/Microsoft.Fx.Portability/Reporting/IReportWriter.cs
```diff
@@ -3,13 +3,14 @@
...
+using System.Threading.Tasks;
 
 namespace Microsoft.Fx.Portability.Reporting
 {
     public interface IReportWriter
     {
         ResultFormatInformation Format { get; }
 
-        void WriteStream(Stream stream, AnalyzeResponse response);
+        Task WriteStreamAsync(Stream stream, AnalyzeResponse response);
```
marklio: Oh, I see. Some of those other async changes were to support this interface change.

tests/lib/Microsoft.Fx.Portability.Cci.Tests/TestAssembly.cs
```diff
@@ -17,7 +17,15 @@ namespace Microsoft.Fx.Portability.Cci.Tests
...
         private const string TFM = @"[assembly: global::System.Runtime.Versioning.TargetFrameworkAttribute("".NETFramework,Version=v4.5.1"", FrameworkDisplayName = "".NET Framework 4.5.1"")]";
-        private static readonly string Mscorlib = typeof(object).GetTypeInfo().Assembly.Location;
+        private static readonly IEnumerable<string> DefaultReferences = new[]
+        {
+            typeof(object).Assembly.Location,
+            typeof(Console).Assembly.Location,
+
+            // Reference to System.Runtime.dll rather than System.Private.CoreLib on .NET Core
+            typeof(RuntimeReflectionExtensions).Assembly.Location
+        }.Distinct();
```
marklio: This is a fragile way to do this. The location of types isn't set, particularly in Core. If we want to be explicit about references, we should just build the explicit list.

marklio: I could be persuaded that this is a nice to have.

Lxiamail: Created a new issue to track the concern, but don't block the integration.

src/lib/Microsoft.Fx.Portability.Reports.DGML/DGMLOutputWriter.cs
Lxiamail: It was for IReportWriter.WriteStream to IReportWriter.WriteStreamAsync

src/lib/Microsoft.Fx.Portability.Reports.Json/JsonReportWriter.cs
Lxiamail: It was for IReportWriter.WriteStream to IReportWriter.WriteStreamAsync

src/lib/Microsoft.Fx.Portability/Utils/JsonConverters/JsonMultiDictionaryConverter.cs
```diff
@@ -9,7 +9,7 @@
 namespace Microsoft.Fx.Portability.Utils.JsonConverters
 {
     /// <summary>
-    /// Json.NET does not handle IDictionary<TKey,TValue> when TKey is a type other than string.  This provides
+    /// Json.NET does not handle <see cref="IDictionary{TKey, TValue}"/> when TKey is a type other than string.  This provides
```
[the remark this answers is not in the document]
Lxiamail: fixed by PR https://github.com/microsoft/dotnet-apiport/pull/815

src/lib/Microsoft.Fx.Portability/AssemblyFileComparer.cs
```diff
@@ -21,7 +22,16 @@ public int Compare(IAssemblyFile x, IAssemblyFile y)
...
-            return string.Compare(x.Name, y?.Name, StringComparison.Ordinal);
+            // Filenames are case insensitive on Windows.
+            var comparison = StringComparison.OrdinalIgnoreCase;
+
+#if !NET461
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                comparison = StringComparison.Ordinal;
+            }
+#endif
```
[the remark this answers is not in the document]
Lxiamail: Tracked by issue https://github.com/microsoft/dotnet-apiport/issues/817, don't block integration

Issue #813, opened 2019-10-30
DGML reference assembly APICounts issues


Lxiamail:
1. When a note.UsageData is null, the remaining nodes in the reference graph was skipped. Reference to Marklio's CR comment at PR https://github.com/microsoft/dotnet-apiport/pull/809.
2. The _searchInGraph reset logic is wrong. It only detects cycle reference, but duplicates counting when a node shows up in more than one branches in the reference graph.

terrajobst:
Closing as API Port was deprecated in favor of binary analysis in [.NET Upgrade Assistant](https://github.com/dotnet/upgrade-assistant/blob/main/docs/binary_analysis.md).

Issue #814, opened 2019-10-30
DGMLOutputWriter merges more than one requests' output together 


Lxiamail:
DGMLOutputWriter.dgml is readonly instance, which only created at by DGMLOutputWriter constructor. However, DGMLOutputWriter instance is shared across requests. This results with DGML output merged more than one requests' output together unexpectedly.

PR #815, opened 2019-10-30
Fixed issue #813 and #814 in DGML format

Lxiamail:
The issues are fixed:

- DGMLOutputWriter merges more than one requests' output together #814
- DGML reference assembly APICounts issues #813

### DIFF
--- a/src/lib/Microsoft.Fx.Portability.Reports.DGML/ReferenceNode.cs
+++ b/src/lib/Microsoft.Fx.Portability.Reports.DGML/ReferenceNode.cs
@@ -10,8 +10,6 @@ namespace Microsoft.Fx.Portability.Reports.DGML
 {
     internal class ReferenceNode
     {
-        private bool _searchInGraph = false;
-
         public List<TargetUsageInfo> UsageData { get; set; }
 
         public string Assembly { get; }
@@ -66,11 +64,11 @@ namespace Microsoft.Fx.Portability.Reports.DGML
             if (Nodes.Count == 0)
                 return 1;
 
-            // reset all the nodes in the reference graph as not be searched.
-            ResetReferenceSearchGraph();
+            // HashSet keep track of the node went through to prevent duplicated counting in the reference graph and circular reference
+            HashSet<ReferenceNode> searchedNodes = new HashSet<ReferenceNode>();
 
             // sum up the number of calls to available APIs and the ones for not available APIs for references.
-            GetAPICountFromReferences(target, out int availableApis, out int unavailableApis);
+            GetAPICountFromReferences(target, searchedNodes, out int availableApis, out int unavailableApis);
 
             // prevent Div/0. When availableApis is 0 and unavailableApis is 0, it likely means that it's references are all not resulted assemblies, return 0.
             if (availableApis == 0 && unavailableApis == 0)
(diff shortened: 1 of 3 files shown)


src/lib/Microsoft.Fx.Portability.Reports.DGML/ReferenceNode.cs
```diff
@@ -66,65 +66,65 @@ public double GetPortabilityIndexForReferences(int target)
...
+                // because we want to only count a reference node once if it is referenced in the graph more than once.
+                if (item._searchInGraph == true)
                 {
-                    // skip the Node with no UsageData
                     continue;
                 }
-
-                if (!item.TryGetAPICountFromReferences(target, out int refCountAvail, out int refCountUnavail))
+                else
                 {
```
marklio: Super NIT: I don't like the "else" here, because technically, the rest of the foreach loop is the else. This is more personal code philosophy though, so it's not a big deal.

src/lib/Microsoft.Fx.Portability.Reports.DGML/ReferenceNode.cs
```diff
@@ -66,65 +66,65 @@ public double GetPortabilityIndexForReferences(int target)
...
-                    // Cycle!
-                    _searchInGraph = false; // Reset this flag
+                    item._searchInGraph = true;
+                }
 
-                    return false;
+                if (item.UsageData != null)
+                {
+                    availAPIs += item.UsageData[target].GetAvailableAPICalls();
```
marklio: Can any of the rest of these calls fail? (can [target] fail with index out of range or another exception, or return null?

src/lib/Microsoft.Fx.Portability.Reports.DGML/ReferenceNode.cs
```diff
@@ -66,65 +66,65 @@ public double GetPortabilityIndexForReferences(int target)
...
+        /// </summary>
+        private void ResetReferenceSearchGraph()
+        {
+            // if we don't have any outgoing references, done reset with this node
+            if (Nodes.Count == 0)
+                return;
+            foreach (var item in Nodes)
+            {
+                item._searchInGraph = false;
+                item.ResetReferenceSearchGraph();
```
marklio: How do you avoid circular references here?

marklio: For what it's worth, my strategy for avoiding following cycles in graphs is to put each node in a HashSet<T> and if Add returns true, process it, otherwise, I've seen it before. It works well for graphs that are reasonable in size, and it doesn't  come with any interesting cleanup problems like this. You can also have multiple threads inspecting the same graph without colliding on state.

marklio: If you wanted a fix for this approach, ResetReferenceSearchGraph could treat _searchInGraph with the oppose rule (only follow Nodes if _searchInGraph was false).

src/lib/Microsoft.Fx.Portability.Reports.DGML/ReferenceNode.cs
Lxiamail: It's gone in updated code.

src/lib/Microsoft.Fx.Portability.Reports.DGML/ReferenceNode.cs
Lxiamail: Great suggestion! Pushed a commit to use the suggested HashSet approach.

src/lib/Microsoft.Fx.Portability.Reports.DGML/ReferenceNode.cs
Lxiamail: Fixed in the new commit.

marklio: Did this actually get fixed? Or we just proved these calls would always succeed?

Lxiamail: Fixed by adding try-catch around item.UsageData[target] statements in the last commit. Any additional concern?

Issue #821, opened 2019-11-08
PublicKeyToken parsing doesn't handle not strong named assembly


Lxiamail:
PublicKeyToken.ParseString() throws exception when the hex string is null.

terrajobst:
Closing as API Port was deprecated in favor of binary analysis in [.NET Upgrade Assistant](https://github.com/dotnet/upgrade-assistant/blob/main/docs/binary_analysis.md).

PR #822, opened 2019-11-08
Fix issue #821: PublicKeyToken parsing doesn't handle not strong name…

Lxiamail:
…d assembly correctly

### DIFF
--- a/src/lib/Microsoft.Fx.Portability/Analyzer/PublicKeyToken.cs
+++ b/src/lib/Microsoft.Fx.Portability/Analyzer/PublicKeyToken.cs
@@ -86,7 +86,7 @@ namespace Microsoft.Fx.Portability
 
         private static ImmutableArray<byte> ParseString(string hex)
         {
-            if (string.Compare(hex, "null", true, CultureInfo.InvariantCulture) ==0)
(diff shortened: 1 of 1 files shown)


src/lib/Microsoft.Fx.Portability/Analyzer/PublicKeyToken.cs
```diff
@@ -83,6 +86,9 @@ public override string ToString()
 
         private static ImmutableArray<byte> ParseString(string hex)
         {
+            if (string.Compare(hex, "null", true, CultureInfo.InvariantCulture) ==0)
+                return ImmutableArray<byte>.Empty;
+
             if (hex.Length % 2 != 0)
```
marklio: Do we also need to check for "actual null" here? Perhaps that should just throw to make diagnostics simpler? Otherwise, we'll nullref here.

Lxiamail: Looks like the serializer always makes "null" as the string token value when the assembly is not strong named, I added the "actual null" check to return ImmutableArray<byte>.Empty.

PR #823, opened 2019-11-12
Fix the issue #821: PublicKeyToken parsing doesn't handle not strong named assembly

Lxiamail:


Lxiamail:
The fix has been reviewed and dotnet-apiport-Linux-CI passed rerun.

PR #828, opened 2019-11-20
Update vsix manifest

Lxiamail:
Update Product ID, version and more info URL for VSIX since a unique ID is required for a new release in market place.

### DIFF
--- a/src/ApiPort/ApiPort.Vsix/source.extension.vsixmanifest
+++ b/src/ApiPort/ApiPort.Vsix/source.extension.vsixmanifest
@@ -1,14 +1,15 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
     <Metadata>
-        <Identity Id="55d15546-28ca-40dc-af23-dfa503e9c5fe" Version="1.4.0" Language="en-US" Publisher="Microsoft" />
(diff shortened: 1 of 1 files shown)


src/ApiPort/ApiPort.Vsix/source.extension.vsixmanifest
```diff
@@ -1,14 +1,15 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
     <Metadata>
-        <Identity Id="55d15546-28ca-40dc-af23-dfa503e9c5fe" Version="1.4.0" Language="en-US" Publisher="Microsoft" />
```
marklio: Do we really have to change the Id? I would assume that this would make it a completely **_different_** extension from the extension manager's perspective rather than simply a new version of an existing extension.

Lxiamail: I got an error when I upload the VSIX to VS marketplace complaining about each extension needs an unique ID. I guess Product Id is what it refers to, but I'm not exactly sure.

Lxiamail: Looks like only version needs to be changed per https://docs.microsoft.com/en-us/visualstudio/extensibility/how-to-update-a-visual-studio-extension?view=vs-2019.

Lxiamail: Unzipped the VSIX file and looked into the manifest file, the version number has been changed as part of the new build. No need to make manual change. Looks like I need to do something with the marketplace publish site.

PR #834, opened 2019-11-26
Add an after build Target to sign the apiport*.nupkg file

Lxiamail:


src/ApiPort/ApiPort/ApiPort.props
```diff
@@ -71,4 +71,12 @@
   <Target Name="AfterBuild">
     <Copy SourceFiles="@(BinPlaceContent)" DestinationFiles="@(BinPlaceContent->'$(TargetDir)\%(Filename)%(Extension)')" SkipUnchangedFiles="true" />
   </Target>
+
+  <Target Name="SignOutputPackage" BeforeTargets="AfterBuild">
```
brianrob: Just confirming that you want this to run before the AfterBuild target.  It isn't 100% clear to me whether you want this to run before that target or after.

Lxiamail: @brianrob good catch! I meant after the AfterBuild target. Just pushed a commit to fix it.

brianrob: Awesome.  Otherwise, LGTM, so feel free to merge when you're ready.

PR #844, opened 2019-12-16
Avoid Invoke-WebRequest for much faster download of catalog.bin

AArnott:
Seriously: this drops the download time from 15+ seconds to <2 seconds.


AArnott:
I'll update this pr

AArnott:
This is ready to review.

PR #843, opened 2019-12-16
Add package dependency to Microsoft.VisualStudioEng.MicroBuild.Core t…

Lxiamail:
…o solve apiport nuget package signing issue

PR #846, opened 2019-12-16
Merge #843: add dependency to Microsoft.visualstudioEng.Microbuild.Core 

Lxiamail:
for fixing the ApiPort NuGet package signing issue

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
-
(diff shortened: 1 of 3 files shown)


AArnott:
Merely adding this won't cause the binaries *inside* the package to be signed. It will only sign the package itself.

To sign everything in the package as well, the easiest way is to simply add a PackageReference to [`Microbuild.VisualStudio`](https://dev.azure.com/azure-public/vside/_packaging?_a=package&feed=vs-impl&package=MicroBuild.VisualStudio&protocolType=NuGet&version=2.0.61&view=overview). If you need to customize the public key token you can do that too.

Lxiamail:
@AArnott Thanks for the comment. The current issue is that APIPort Nuget package is not signed when it is built as global tool. Hope this change help.

AArnott:
> Hope this change help.

FYI you can *fully* test this locally on a dev box using test signing. No reason to check in and 'hope' it works. :)

PR #848, opened 2019-12-17
Switch from GitVersion to Nerdbank.GitVersioning

AArnott:
This is .NET Core compatible so it runs on multiple OS's, doesn't require a restore during the build as your GitVersion use did, and it doesn't require a special azure pipelines step.
It sets the build number in Azure Pipelines to match the version actually built.


AArnott:
This was done as part of #869 in order to make the build more reliable, since GitVersion was failing the build too frequently.

PR #849, opened 2019-12-20
Make apiport package id starts with "Microsoft"

Lxiamail:
to follow Microsoft released tool guidance at https://docs.microsoft.com/en-us/dotnet/core/tools/troubleshoot-usage-issues#package-naming-enforcement.
Also remove the target "GetSignNuGetPackFiles", which is not needed.

### DIFF
--- a/src/ApiPort/ApiPort/ApiPort.props
+++ b/src/ApiPort/ApiPort/ApiPort.props
@@ -14,6 +14,7 @@
         <TargetFramework>netcoreapp2.1</TargetFramework>
         <PackAsTool>true</PackAsTool>
         <ToolCommandName>ApiPort</ToolCommandName>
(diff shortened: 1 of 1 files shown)


src/ApiPort/ApiPort/ApiPort.props
```diff
@@ -14,6 +14,7 @@
         <TargetFramework>netcoreapp2.1</TargetFramework>
         <PackAsTool>true</PackAsTool>
         <ToolCommandName>ApiPort</ToolCommandName>
+          <PackageId>Microsoft.ApiPort</PackageId>
```
AArnott, line 17: nit: indentation

Lxiamail: Fixed

PR #850, opened 2020-01-07
added analyze.sh to match existing build.sh (w/updated docs)

pracplayopen:
added posix analyze.sh script to simplify generating reports.  also updated docs to reflect


AArnott:
Thanks very much for sharing your pull request.
We're moving in a different direction at this point. We actually removed the build.sh file in favor of standardizing on PowerShell Core scripts which run cross platform.

PR #860, opened 2020-03-31
Require v2.2 in build lab as well to fix build error

Lxiamail:
Without specify the .NET Core version to use, in build lab with VS2019, 3.0 is used by default, we got the following build error. Using .NET Core 2.2 avoid these build errors
C:\Program Files\dotnet\sdk\3.1.200\Sdks\Microsoft.NET.Sdk.Razor\build\netstandard2.0\Microsoft.NET.Sdk.Razor.Compilation.targets(130,10): Error MSB4064: The "DisableSdkPath" parameter is not supported by the "Csc" task. Verify the parameter exists on the task, and it is a settable public instance property.

C:\Program Files\dotnet\sdk\3.1.200\Sdks\Microsoft.NET.Sdk.Razor\build\netstandard2.0\Microsoft.NET.Sdk.Razor.Compilation.targets(114,5): Error MSB4063: The "Csc" task could not be initialized with its input parameters.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,3 +1,6 @@
 {
-    "sources": [ "src" ]
+    "sources": [ "src" ],
+"sdk": {
+  "version": "2.2.401"
+}
 }


Lxiamail:
Request SDK v2.2 doesn't work in the build pipeline. It failed at SDK installation time due to not be able to find the specified SDK version. Instead of doing this, Switched back to VS2107 for now until moving ApiPort code to .NET Core v3.*

Lxiamail:
> I'm not familiar with the build process, but this seems reasonable. Is there a reason we're not on 3.x?

Since asp.net core 3.x changed quite a bit from v2.2, we should move to 3.x, just we need to work on other high priority items before getting to it.

marklio:
You should be able to use the 3.0 SDK but still target netstandard or core 2, or whatever.

Lxiamail:
> You should be able to use the 3.0 SDK but still target netstandard or core 2, or whatever.

Yes, however, we ran into the build errors as stated in the issue's description.

PR #862, opened 2020-04-04
Fix accessibility bug#1059520

Lxiamail:
Fix bug https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1059520


Lxiamail:
> If I understand, this is just reducing confusion about the expectation of what the button does. Since it doesn't navigate to a different location, it should be a button, not a "hyperlink". Seems reasonable.

Yes, that's my understanding as well.

PR #865, opened 2020-04-11
Fix bug#945642 by removing NETSTANDARD1_3 complication condition. 

Lxiamail:
We can do the fix now  since NetStandard1.3 had been removed from support TFM list in our project files.


Lxiamail:
@marklio Thanks!

PR #869, opened 2020-04-18
Add azure-pipelines.yml file

AArnott:
This is almost exactly equivalent to the GUI-based pipeline you have now. There are lots of improvements I'd like to make but I'll save those for other PRs so you can review them individually.

You can see the pipeline file working here: https://dev.azure.com/devdiv/Personal/_build?definitionId=12905&_a=summary

Once you accept this PR, you should edit or recreate your official Azure Pipeline to use this yml file instead of the GUI-based one.

* [x] dotnet-apiport-dev-CI recreated as YAML
* [x] dotnet-apiport-linux-CI recreated as YAML

Note that `dotnet-apiport-linux-CI` wasn't actually Linux. It was trying (and failing) to use Mac. In my recreation of it, I actually used Linux since that works, and fits the name.


azure-pipelines.yml
```diff
@@ -0,0 +1,209 @@
...
+  - task: securedevelopmentteam.vss-secure-development-tools.build-task-publishsecurityanalysislogs.PublishSecurityAnalysisLogs@2
+    displayName: 'Publish Security Analysis Logs'
+    condition: succeededOrFailed()
+
+- job: Linux
+  pool:
+    vmImage: Ubuntu 18.04
+  variables:
+    BuildNetCoreAppTarget: netcoreapp2.1
+    RestoreBuildProjects: src/ApiPort/ApiPort/*.csproj
```
Lxiamail, line 171: Do we need to restore the lib projects under src\lib\**\*.csproj here?

azure-pipelines.yml
```diff
@@ -0,0 +1,209 @@
...
+  - task: DotNetCoreCLI@2
+    displayName: dotnet publish
+    inputs:
+      command: publish
+      publishWebProjects: false
+      projects: $(RestoreBuildProjects)
+      arguments: '--configuration $(BuildConfiguration) -f $(BuildNetCoreAppTarget) --output $(build.artifactstagingdirectory) --no-build'
+      zipAfterPublish: True
+
+  - publish: $(build.artifactstagingdirectory)
```
Lxiamail, line 208: previous UI pipeline definition has artifact name as "drop". Should we define it too here, or it is default value anyway?

azure-pipelines.yml
Lxiamail: I looked into current dotnet-ApiPort-Linux-CI definition, It only has "src/ApiPort/ApiPort/*.csproj", I guess that you are just doing what it was doing here. I haven't worked on Linux build definition before, I'm not sure if it was broken in the UI definition always. Anyway, we can keep it as is, and test it out later.

global.json
```diff
@@ -1,3 +1,7 @@
 {
-    "sources": [ "src" ]
-}
\ No newline at end of file
+    "sdk": {
+      "version": "2.1.513",
+      "rollForward": "patch",
```
Lxiamail, line 4: Do you know if v2.1.513 SDK is available in Azure DevOp? I tried to specify v2.2.401 previously, but it fails in the build lab with the error of not finding the specified version of SDK.

azure-pipelines.yml
```diff
@@ -0,0 +1,209 @@
+trigger:
+  branches:
+    include:
+    - master
```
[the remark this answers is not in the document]
AArnott: That might depend on just how different your pipelines would be between the two branches. But maintaining multiple pipelines tends to be more work, and usually they have quite a bit in common. So I would recommend we start with one pipeline that handles both master and dev, and we add conditions or variables as necessary to get whatever distinct behavior you want between the two branches.

azure-pipelines.yml
```diff
@@ -0,0 +1,209 @@
...
+    displayName: Copy Analysis files to Staging
+    inputs:
+      SourceFolder: bin\$(BuildConfiguration)
+      Contents: |
+        ApiPort\**\*
+        ApiPort.Offline\**\*
+        ApiPort.Vsix\*.vsix
+        raw\**\*
+        *.binlog
+      TargetFolder: $(Build.StagingDirectory)\drop
```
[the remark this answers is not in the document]
AArnott: No. Having an artifact name that is inconsistent from build to build makes for harder to use artifacts. I changed it from `$(Build.BuildNumber)` to just `drop` so that it was the same across builds.

azure-pipelines.yml
```diff
@@ -0,0 +1,209 @@
...
+  - publish: $(Build.StagingDirectory)\drop
+    displayName: Publish drop
+    artifact: drop
+    enabled: false
+
+  - task: securedevelopmentteam.vss-secure-development-tools.build-task-fxcop.FxCop@2
+    displayName: 'Run FxCop'
+    inputs:
+      inputType: Basic
+      targets: 'bin\$(BuildConfiguration)\ApiPort*\net4*\**\ApiPort*.*;bin\$(BuildConfiguration)\ApiPort.Offline\net4*\**\Microsoft.Fx.*.dll;'
```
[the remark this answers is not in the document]
AArnott: I didn't make this up. This came directly from your existing pipeline, so I can't say why it was this way.
That said, in my experience FxCop fails miserably on anything other than .NET Framework targeted assemblies. So I can imagine why it was done this way. 
Further: I recommend we drop FxCop entirely since it's a deprecated thing. We should switch to the analyzer alternative (in a subsequent PR).

azure-pipelines.yml
```diff
@@ -0,0 +1,209 @@
...
+      recursive: false
+      directory: 'C:\Windows\Microsoft.NET\Framework\v4.0.30319'
+      verbose: true
+    enabled: false
```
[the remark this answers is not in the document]
AArnott: Not in this PR. It is disabled in your pipeline already. The point of this PR is to reproduce what you already had, but in YAML form.

azure-pipelines.yml
AArnott: Yup: there are oh-so-many improvements/fixes I'd like to make, but I'm holding off so that we can review each of those changes individually in PRs. If I made a bunch of changes here as part of this PR, it would be hard to know what I *changed* vs. what I reproduced from your existing pipelines.

azure-pipelines.yml
AArnott: The default value is "Linux", I guess named after the job name.
The original linux pipeline used `drop` as an artifact name. 

We could name this one `drop-Linux` and the Windows one `drop-Windows` if you'd like.

global.json
AArnott: Yes, VS 2017 agents have 2.1.513 installed. I used this version because this seems to be what you were *already* using, although it wasn't explicit. You were just using whatever was on the agent. But this was a problem because your build _breaks_ on newer .NET SDK versions, so as a step toward stable builds on all machines, and eventually moving to newer build agents, we need to 'pin' what you depend on so we can intentionally upgrade when we are ready with a build that is tested to work on that version.

But we should *not* depend on agents to have this installed. I would like to follow up after this PR with another PR that will install the .NET SDK specified in this file onto the agent, so that we no longer are at the mercy of the agent image to have the prereq that we need.

AArnott: As evidence that this works as-is, you can see my build based on this PR: https://dev.azure.com/devdiv/Personal/_build/results?buildId=3654797&view=results

azure-pipelines.yml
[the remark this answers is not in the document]
AArnott: I just took a look at your `dotnet-apiport-master-Release` pipeline. It seems to me that it's distinct from the dev-CI pipeline in that it's meant for *releases* rather than a per-branch difference. 
So building master with this new YML file should be fine. But as a follow-up to this PR, we can take another look at the master-Release pipeline and integrate its unique abilities into this YAML so that we can build releases as well as handle PR/CI builds and eventually retire the master-Release pipeline too.

azure-pipelines.yml
AArnott: I've renamed the drops as I proposed.

azure-pipelines.yml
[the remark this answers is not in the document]
Lxiamail: Sounds good plan

azure-pipelines.yml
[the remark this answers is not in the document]
Lxiamail: In the current pipeline definition in UI, it has "Run Roslyn Analyzers" task and enabled. Is it the Roslyn analyzer that we should use?

Lxiamail:
The [current UI build pipleline definition](https://devdiv.visualstudio.com/DevDiv/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=7913) has "Component detection" task, looks like it is missing from this PR.

AArnott:
The "Component detection" task is not missing. It's [right here](https://github.com/microsoft/dotnet-apiport/pull/869/files#diff-fec826feae04e51c0d94076385408bdcR126-R127).

azure-pipelines.yml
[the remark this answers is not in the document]
AArnott: I've never seen that task before, and it repeats a *lot* of work your pipeline already did so I'd love to remove it. 
The analyzers I'd like to see added so we can remove this fxcop task is [this set](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers/2.9.8).

AArnott:
Thanks for merging this. I've created https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?definitionId=12912&_a=summary as the new YAML based pipeline.
I suggest you delete https://dev.azure.com/devdiv/DevDiv/_build?definitionId=7913&_a=summary once you see a good build from the new one.

PR #875, opened 2020-04-29
Integration from Dev to master with some bug fixes and a bunch of infrastructure work

Lxiamail:
primarily include:
- Fixed a race condition in DGML writer
- Fixed 2 accessibility bugs
- Fixed an code analysis bug
- Updated TFM to support .NET Core 3.1
- Used a unique Guid for the VS option page to avoid collision with other VS extensions
- Enabled YAML build pipeline and some build pipeline clean up
- Installed SDK specified in global.json so that dev and azure DevOps environments using consistent SDK version.
- Switched from GitVersion to Nerdbank.GitVersioning
- Updated APIPort global tool package ID to start with "Microsoft" prefix


AArnott:
This is fine, but I wonder if this merge will work with the existing master-release pipeline. 
I had in mind to integrate what was unique about the master-release pipeline into the yml file that no provides the main CI/PR for the dev branch so that all flows including releasing can be done out of a source controlled yml file. I guess you can just merge that in a separate PR into master later. It's up to you.

Lxiamail:
> This is fine, but I wonder if this merge will work with the existing master-release pipeline.
> I had in mind to integrate what was unique about the master-release pipeline into the yml file that no provides the main CI/PR for the dev branch so that all flows including releasing can be done out of a source controlled yml file. I guess you can just merge that in a separate PR into master later. It's up to you.

You are right. Looks like the current yml pipeline doesn't include all the tasks in master. I do want to get an update in master to unblock .NET Portability Service release, which shares the same portability libraries with APIPort. I will go with the current yml file as it is and still leave the master's UI pipeline enabled.

Lxiamail:
Created 2.7.0-alpha.1 tag in master for code prior to this PR.

PR #881, opened 2020-05-04
Fix "Analyze Assembly" to "Analyze Solution"

twsouthwick:
This must have been a copy/paste error when it was put in. This is the selection on the solution node.


src/ApiPort/ApiPort.VisualStudio/ApiPort.VisualStudio.vsct
```diff
@@ -83,7 +83,7 @@
         <CommandFlag>DynamicVisibility</CommandFlag>
         <CommandFlag>DefaultInvisible</CommandFlag>
         <Strings>
-          <ButtonText>Analyze Assembly Portability</ButtonText>
+          <ButtonText>Analyze Solution Portability</ButtonText>
```
Lxiamail, line 86: Looks like this menu button goes to two places, one is under "Analyze" menu, the other is when right click on the solution. This change is applicable well for later, but may not for former because it can point to any assembly, which doesn't have to be the current solution.

Issue #888, opened 2020-06-16
NuGet.config file is overridden by global NuGet.Config


bashenk:
I was unable to get the build working due to missing System.CommandLine, until I added this line to my `<packageSources>` in my NuGet.config:
### Edit 
Turns out the issue was actually that it was using my global NuGet.Config file instead of the one in the build directory. Maybe consider adding a `--configfile ./NuGet.config` argument to the `dotnet restore` command in ./init.ps1

The error I was getting was `ApiPort depends on System.CommandLine (>= 0.1.0-e171125-2) but System.CommandLine 0.1.0-e171125-2 was not found. An approximate best match of System.CommandLine 0.3.0-alpha.20054.1 was resolved`

Lxiamail:
@bashenk Thanks for reporting the issue. Since you already figured out a fix, do you like to submit a PR?

bashenk:
@Lxiamail Sure, I should be able to submit one tomorrow.

PR #889, opened 2020-06-17
Rename NuGet.config to NuGet.Config

bashenk:
Nuget fails to find the solution's config file on Windows when the solution is in a case-sensitive folder unless the file name is exactly "NuGet.Config"

Closes issue #888


bashenk:
@Lxiamail Okay I've pushed an update. Are there any additional issues?

PR #890, opened 2020-06-18
Update pipeline to use APIScan v3

Lxiamail:


Lxiamail:
APIScan v3 is required in Azure DevOps pipeline

PR #891, opened 2020-06-18
Added exception info capabilities into apiport

LoopedBard3:
Added models and methods to allow for the finding of APIs that may throw exceptions in addition to the general catalog API information. A majority of the changes are in the portability libraries because that is what does a majority of the processing and what holds the API lookup class. Also updated the Excel writer to all for the writing of the exception information.


LoopedBard3:
Separating the PR's into library and apiport PR's

PR #898, opened 2020-07-16
Update azure-pipelines.yml

LoopedBard3:
Add a condition to the nuget command so that it only trys to upload on dev builds

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -143,6 +143,7 @@ jobs:
     condition: succeededOrFailed()
 
   - task: NuGetCommand@2
+    condition: eq(variables['Build.SourceBranch'], 'refs/heads/dev')
     inputs:
       command: 'push'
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -143,6 +143,7 @@ jobs:
     condition: succeededOrFailed()
 
   - task: NuGetCommand@2
+    condition: eq(variables['Build.SourceBranch'], 'refs/heads/dev')
```
Lxiamail, line 146: Master branch needs to push nuget packages as well. Can you distingish them by different versions to avoid confliction?

Lxiamail: Ha, I understand the goal of the PR better now. The goal is not to void version conflict with master branch, but avoid push nuget package in CI build since the CI build fails for pushing nuget package due to access issue. With that background, let's including master in this condition so that we push nuget package in master build as well.

PR #901, opened 2020-07-30
Added the retrieval of the exceptions.bin file to the init script.

LoopedBard3:
Added the retrieval of the exceptions.bin file to the init script. It is setup to use the same storage as the catalog.bin file. However, the exceptions file is not there yet so a warning will be thrown until the file is added, but I tested with a dev blob storage to make sure it works in general and it did download.

**This should not be merged until the exceptions file is added to the portability storage but will only show warnings if used before then.**

### DIFF
--- a/tools/Get-AdditionalDataBin.ps1
+++ b/tools/Get-AdditionalDataBin.ps1
@@ -1,0 +1,33 @@
+$ErrorActionPreference = "Stop"
+
+function DownloadFile($url, $outputPath) {
+	Write-Host "Attempt to download to $outputPath"
+
+	# If the file has been downloaded don't download again. An empty file implies a failed download
+	if(Test-Path $outputPath) {
+		$file = Get-ChildItem $outputPath
+
+		if($file.Length -gt 0) {
+			Write-Host "$outputPath is already downloaded"
+			return;
+		}
+	}
+
+	try {
+		# Create placeholder so directory exists
+		New-Item -Type File $OutputPath -Force | Out-Null
+
+		# Attempt to download.  If fails, placeholder remains so msbuild won't complain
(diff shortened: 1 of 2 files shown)


twsouthwick:
It would be great for this to be retrieved via the portability service so no one has to have access to the blob storage.

Same thing for the catalog.bin file.

@Lxiamail @marklio

Lxiamail:
> It would be great for this to be retrieved via the portability service so no one has to have access to the blob storage.
> 
> Same thing for the catalog.bin file.
> 
> @Lxiamail @marklio

Yes, portability service directly load it from blob storage. I think Tayler's change here is making the parity change as catalog.bin download functionality in init file for the local usage scenario.

twsouthwick:
My comment probably shouldn't block this PR (it would require updating the service to load them). It was more a suggestion for future work to serve the data via ApiPort service rather than directly from blob storage

tools/Get-AdditionalDataBin.ps1
```diff
@@ -0,0 +1,33 @@
+$ErrorActionPreference = "Stop"
+
+function DownloadFile($url, $outputPath) {
```
twsouthwick, line 3: Doesn't the `Get-CatalogBin.ps1` script have a similar method? Maybe combine the two as `Get-DataFiles.ps1` so there's no duplication

Lxiamail: Sounds good idea. I will merge this PR once we have the exception.bin file in the blob storage, then make a new PR to merge the duplicated code.

terrajobst:
Closing as API Port was deprecated in favor of binary analysis in [.NET Upgrade Assistant](https://github.com/dotnet/upgrade-assistant/blob/main/docs/binary_analysis.md).

PR #904, opened 2020-08-03
Remove .NET Framework build of ApiPort

twsouthwick:
.NET Core 3.1 provides the APIs for web proxy that was the last remaining difference with .NET Framework. This change removes the .NET Framework build of ApiPort and enables proxy support using HttpClient.GetDefaultProxy. As part of this change, it simplifies the global generation as well.


azure-pipelines.yml
```diff
@@ -129,15 +114,6 @@ jobs:
       AnalyzeSymPath: $(Build.StagingDirectory)\symbols
     continueOnError: true
 
-  - task: securedevelopmentteam.vss-secure-development-tools.build-task-apiscan.APIScan@2
```
Lxiamail, line 132: Should we run APIScan against v3.1 binaries instead of moving it?

twsouthwick: Put it back

PR #918, opened 2020-09-01
Change to MemoryStream.CopyToAsync()

ALCoder6:
ASP.NET Core 3.1 disabled sync IO operation by default.

### DIFF
--- a/src/lib/Microsoft.Fx.Portability.Reports.DGML/DGMLManager.cs
+++ b/src/lib/Microsoft.Fx.Portability.Reports.DGML/DGMLManager.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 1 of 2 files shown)


src/lib/Microsoft.Fx.Portability.Reports.DGML/DGMLManager.cs
```diff
@@ -120,7 +120,7 @@ internal void Save(Stream stream)
             {
                 file.Save(ms);
                 ms.Position = 0;
-                ms.CopyTo(stream);
+                ms.CopyToAsync(stream);
```
twsouthwick, line 123: The entire chain will need to be async. Can't just change it here

Lxiamail: Fixed

PR #924, opened 2020-09-21
Add unique tasks for master branch 

Lxiamail:
so that we can switch master branch to use this pipeline.


azure-pipelines.yml
```diff
@@ -136,6 +153,23 @@ jobs:
...
+    inputs:
+      scriptType: inlineScript
+      workingFolder: 'bin\$(BuildConfiguration)'
+      inlineScript: |
+        # Setting these environment variables because they are used in the script for the upload endpoint: 
+     
+        $env:APPVEYOR_REPO_PROVIDER = "github"
+        $env:APPVEYOR_REPO_NAME = "microsoft/dotnet-apiport"
+     
+        Invoke-WebRequest "https://raw.github.com/madskristensen/ExtensionScripts/master/AppVeyor/vsix.ps1" -UseBasicParsing | Invoke-Expression
```
marklio, line 168: What does this script do? Mads is clearly a reasonable person, but this seems like a liability in an official build. Should we host our own copy of this script?

Lxiamail: Good point. However, I hope there is more official channel hosting the same script instead maintaining our own copy. Following with with Mads to get more advice.

Lxiamail: There is no other more official channel to get the script other than https://raw.github.com/madskristensen/. Create local script with subset of only what we need.

azure-pipelines.yml
```diff
@@ -136,6 +153,22 @@ jobs:
...
+  - task: PowerShell@1
+    condition: eq(variables['Build.SourceBranch'], 'refs/heads/master')
+    displayName: 'Push VSIX'
+    inputs:
+      scriptType: inlineScript
+      inlineScript: |
+        # Setting these environment variables because they are used in the script for the upload endpoint: 
+     
+        $env:APPVEYOR_REPO_PROVIDER = "github"
+        $env:APPVEYOR_REPO_NAME = "microsoft/dotnet-apiport"
```
AArnott, line 165: Would this be a good time to rename these env vars to something more current? 
In fact, you don't actually need them at all, since both of these values are [predefined variables](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml).

Old | New
--|--
`$env:APPVEYOR_REPO_PROVIDER` | `$(Build.Repository.Provider)`
`$env:APPVEYOR_REPO_NAME` | `$(Build.Repository.Name)`

azure-pipelines.yml
```diff
@@ -136,6 +153,22 @@ jobs:
...
+     
+        "tools\AppVeyor\vsix.ps1" | Invoke-Expression
+        $vsix = gci -path bin\$(BuildConfiguration) -Recurse -Include ApiPort.vsix -File
+        $vsix
+        Vsix-PublishToGallery -path $vsix
```
AArnott, line 170: I would not publish to an external gallery directly from my build pipeline. Instead, I recommend you define a second YAML file that consumes the output of this one. See [this example](https://github.com/AArnott/Library.Template/blob/microbuild/azure-pipelines/release.yml) of a release pipeline that consumes the build pipeline.
You can even create a trigger so that whenever you push to master the release pipeline is triggered, thereby removing extra conditions and simplifying your regular build pipeline.

Lxiamail: @AArnott Thanks for your feedbacks! Submitted a new commit with
1. Created a separated release pipleline.
2. Moved common code for both pipeline into two *common*.yml files and used them as template files.

azure-pipelines.yml
Lxiamail: Good to know. Fixed

azure-pipelines.yml
```diff
@@ -122,6 +134,11 @@ jobs:
       softwareVersionNum: 4.XX
       symbolsFolder: '$(Build.StagingDirectory)\drop\raw\symbols\netcoreapp3.1'
       isLargeApp: false
+
+  - task: securedevelopmentteam.vss-secure-development-tools.build-task-roslynanalyzers.RoslynAnalyzers@2
+    displayName: 'Run Roslyn Analyzers'
+    inputs:
+      suppressionFileForCompilerWarnings: 'bin\RoslynAnalyzerSuppressions.txt'
```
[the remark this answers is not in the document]
Lxiamail: roslyn analyzer nuget packate actually has been already referenced in Directory.Build.targets file. Removed this task from pipeline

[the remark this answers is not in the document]
Lxiamail: Created release pipeline with real signing task.

azure-pipelines.yml
```diff
@@ -20,6 +21,17 @@ jobs:
   - powershell: .\init.ps1 -UpgradePrerequisites -NoRestore
     displayName: init.ps1 script
 
+  - task: ms-vseng.MicroBuildTasks.30666190-6959-11e5-9f96-f56098202fef.MicroBuildSigningPlugin@2
+    condition: eq(variables['Build.SourceBranch'], 'refs/heads/master')
```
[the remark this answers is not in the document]
Lxiamail: Created release pipeline with real signing task. I would assume that test signing in all the build will take extra time. Is that something worth to concern?

azure-pipelines-release.yml
```diff
@@ -0,0 +1,42 @@
...
+ - task: PowerShell@1
+    condition: eq(variables['Build.SourceBranch'], 'refs/heads/master')
+    displayName: 'Push VSIX'
+    inputs:
+      scriptType: inlineScript
+      inlineScript: |
+        # Setting these environment variables because they are used in the script for the upload endpoint: 
+     
+        $env:APPVEYOR_REPO_PROVIDER = "github"
+        $env:APPVEYOR_REPO_NAME = "microsoft/dotnet-apiport"
```
AArnott, line 34: These env vars aren't used now, are they?
Also it is the only script in a task labeled "Push VSIX", but it doesn't push any VSIX any more. Was it supposed to invoke the `vsix.ps1` script?
I recommend you drop the condition since this pipeline should do its job (when invoked) regardless of what the branch is called. So I think you can replace this whole task with this:

```yaml
  - powershell: tools/appveyor/vsix.ps1
    displayName: Push VSIX
```

AArnott (changes requested):
This looks better. I'm going to push a commit or two to another branch with some more proposed changes. I see some opportunity to consolidate YAML between windows and linux. 

And I generally like to see the release pipeline consume the output of the main pipeline instead of re-execute it, but I'm interested in whether you have a preference there.

azure-pipelines.yml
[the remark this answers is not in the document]
AArnott: Real signing can really slow a build down but test signing doesn't -- it's all locally executed.

Lxiamail:
> And I generally like to see the release pipeline consume the output of the main pipeline instead of re-execute it, but I'm interested in whether you have a preference there.

no, I don't have preference, I rather prefer not reexecute since it's preferable to ship the same set of binaries tested.

azure-pipelines-release.yml
Lxiamail: Accidently get the wrong code here. Actually, it makes sense to have vsix.ps1 does the push as well. I will update a commit later.

Lxiamail: Commited a fix

AArnott:
Please see #926 for the changes I recommend for this PR.

Lxiamail:
@AArnott Is there any additional feedback? I need your approval to merge the PR.

AArnott (approved):

PR #926, opened 2020-09-28
Refactor build pipeline

AArnott:


Lxiamail (commented):
@AArnott Thank you so much for taking time to clean up the Azure pipeline scripts! It's great to have the script more componentized. I have a general question. There are quite a few files are general purpose scripts, which looks valuable to every azure pipeline script, like init.cmd, azure-pipelines/variabels/_all.ps1, _pipelines.ps1, tools/install-donetsdk.ps1, tools/install-nugetcredprovider.ps1, tools/set-envVars.ps1. I'm wondering if there is any central place these scripts are persisted so that each individual repo doesn't need to keep a separated copy.  I noticed that you made some changes to  tools/install-donetsdk.ps1, tools/install-nugetcredprovider.ps1, tools/set-envVars.ps1 in this PR, these files were commited in your previous PR.  Since you are the expert in this space, I guess that you maybe able to easily identify things need to be changed, however, I'm worry about maintaining these scripts may not be easy thing do to for the others.

azure-pipelines/build-windows.yml
```diff
@@ -0,0 +1,90 @@
+steps:
+- checkout: self
+  clean: true
+
+- template: install-dependencies.yml
+
+- powershell: '& (./azure-pipelines/Get-nbgv.ps1) cloud'
+  displayName: Set build number
+
+- ${{ if eq(variables['system.collectionId'], '011b8bdf-6d56-4f87-be0d-0092136884d9') }}:
```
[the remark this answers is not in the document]
AArnott: That is the `devdiv` project collection. AFAIK all your pipelines are defined there so the `if` check always passes. But if you were to expose a pipeline on the `azure-public` account so your OSS contributors can see what goes wrong with their PR builds, then this `if` will evaluate to false, allowing the build to succeed on azure-public.

azure-pipelines/variables/SignType.ps1
```diff
@@ -0,0 +1,11 @@
+if ($env:SYSTEM_COLLECTIONID -eq '011b8bdf-6d56-4f87-be0d-0092136884d9') {
```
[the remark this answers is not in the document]
AArnott: See earlier answer.

AArnott:
I don't know a good way to avoid checking in these boilerplate scripts. Most of them are subject to tinkering for individual repos, so importing them in a way that makes them readonly to you is counter-productive. The few files that tend to be exactly the same across repos are few enough and slow moving enough that I haven't crafted a way to share just that subset.

AArnott:
But the source of truth for these is https://github.com/aarnott/library.template.

Lxiamail:
@AArnott  build failed with error message:  Response status code does not indicate success: 403 (Forbidden - User '6d3b3c1a-123d-454c-a78b-b4a426164711' lacks permission to complete this action. You need to have 'ReadPackages'. (DevOps Activity ID: 2CF1D764-FEDC-446D-BA87-7C1FEC70684D)) at https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4104702&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=92eaab8a-d904-52ba-6bd2-a55f3ebce252&l=36 . Looks like the build passed when you manually queue it. It is the build failure due to PR build doesn't have the right permission to access internal DevOps feed?

AArnott:
That failure is very odd. I've never seen it before, but I believe you're right: secrets are not exposed to PRs from forks as mine is. Once you accept my PR it should work in yours.
However since yours is an OSS repo, you may want to copy another pattern that we use for our OSS repos: skip the microbuild signing step for PRs, or at least PRs from forks. I'll prepare that change.

AArnott:
It's fixed. :)

azure-pipelines/VSMarketplaceTools.ps1
```diff
@@ -6,25 +6,26 @@ $vsixUploadEndpoint = "https://www.vsixgallery.com/api/upload"
 #$vsixUploadEndpoint = "https://localhost:44372/api/upload"
```
Lxiamail, line 6: @AArnott This file is for upload to https://www.vsixgallery.com/.  There is https://marketplace.visualstudio.com/ called VS marketplace. Looks like they are different galleries, right?

AArnott: I've never heard of vsixgallery.com. And notably absent from this script is any mention of a 'secret' or token or anything that would typically be required to upload a VSIX someplace.

Lxiamail: > I've never heard of vsixgallery.com. And notably absent from this script is any mention of a 'secret' or token or anything that would typically be required to upload a VSIX someplace

I happen to know vsixgallery.com by working on this PR. :) Yes, I would expect some kind of 'secret' is required, but apparently, not for vsixgallery.com.

PR #928, opened 2020-10-13
Minor fix to release.yml 

Lxiamail:


Lxiamail:
Minor yml file, skip CR to speed up

PR #930, opened 2020-10-14
Create a azure-pipelines-release.yml for release 

Lxiamail:
Currentlly, all the CI builds are test signed. When it's release time, we need real signed build and publish vsix file to vsix gallery. This will be release pipeline. Only is triggered manually when we are ready to release.

### DIFF
--- a/azure-pipelines-release.yml
+++ b/azure-pipelines-release.yml
@@ -1,0 +1,35 @@
+trigger:
+  branches:
+    include:
+    - master
+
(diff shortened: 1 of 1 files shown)


Lxiamail:
> Seems reasonable. I'm no YAML expert, but if this produces what we're looking for, then it looks fine.

Unfortunately, it seems almost impossible to verify YAML pipeline on local machine. Have to check in and try it out.

Issue #938, opened 2020-10-27
Analyzing projects in VS fails due to file not found


Lxiamail:
This is an issue only in the code not released yet. It's a regression in ProjectAnalyzer.AnalyzeProjectAsync(). targetAssemblies should  full path of the assembly file instead of file name only.

PR #939, opened 2020-10-27
fix issue Analyzing projects in VS fails due to file not found #938

Lxiamail:
fix issue #938

### DIFF
--- a/src/ApiPort/ApiPort.VisualStudio.Common/Analyze/ProjectAnalyzer.cs
+++ b/src/ApiPort/ApiPort.VisualStudio.Common/Analyze/ProjectAnalyzer.cs
@@ -83,7 +83,7 @@ namespace ApiPortVS.Analyze
 
                 foreach (var file in output)
                 {
-                    targetAssemblies.Add(Path.GetFileName(file));
+                    targetAssemblies.Add(file);
                 }
 
                 referencedNuGetPackages.UnionWith(GetPackageReferences(project));


src/ApiPort/ApiPort.VisualStudio.Common/Analyze/ProjectAnalyzer.cs
```diff
@@ -83,7 +83,7 @@ public async Task AnalyzeProjectAsync(ICollection<Project> projects, Project ent
 
                 foreach (var file in output)
                 {
-                    targetAssemblies.Add(Path.GetFileName(file));
+                    targetAssemblies.Add(file);
```
marklio, line 86: Do we expect "file" to be fully-qualified? I'm just thinking of other ways this could fail. Could if ever be relative to something other than the working directory?

marklio (approved):
Looks like a good targeted fix. Just curious if there are other failure modes.

src/ApiPort/ApiPort.VisualStudio.Common/Analyze/ProjectAnalyzer.cs
Lxiamail: Under debugger, I can see the value of file is absolute path. I'm not sure if in anycase it could be different. Anyhow, the current code total is totally broken. This fix definitly makes it better.

Issue #936, opened 2020-10-27
ExcelOpenXmlOutputWriter doesn't handle gracefully when NonUserAssemblies is null


Lxiamail:
This is a bug in not released ExcelOpenXmlOutputWriter code. Since NonUserAssemblies is null in this release. The previously released APIPort client tool doesn't have the capability to collect NonUserAssemblies. In this case, the portabilityservice should detect NonUserAssemblies is null and skip.

PR #940, opened 2020-10-27
Merge regression issue fixes

Lxiamail:
- Analyzing projects in VS fails due to file not found #938
- ExcelOpenXmlOutputWriter doesn't handle gracefully when NonUserAssemblies is null #936


Lxiamail:
Since both fixes have been reviewed in dev branch. Merging without CR

PR #943, opened 2020-10-30
publish everything from the staging directory to artifacts

Lxiamail:
publish everything from the staging directory to artifacts, including binaries, symbols and zip.


Lxiamail:
verifed that with the PR, binaries, symbols and zip files are published to artifacts. ApiPort NuGet package is real signed, APIPort.dll is also signed.,

PR #945, opened 2020-11-04
Update ApiPort from alpha release to stable and some apiport package changes

Lxiamail:
Updated ApiPort from alpha release to stable so that it's easy to install it as global tool.
Changed ApiPort NuGet package ID to be consistent with App name.
Updated copyright
Added a few package properties


marklio (approved):
Looks good. The only feedback was to make the nuspec content look similar to try-convert.

PR #947, opened 2020-11-05
Set PackageRequireLicenseAcceptance to true

Lxiamail:
to be consistent with try-convert tool


Lxiamail:
@marklio We chatted about this change in the previous PR code review, but I fogot to apply before merging.

Directory.Build.props
```diff
@@ -48,7 +48,7 @@
     <Authors>Microsoft</Authors>
     <PackageLicenseFile>LICENSE.txt</PackageLicenseFile>
     <PackageProjectUrl>https://github.com/Microsoft/dotnet-apiport</PackageProjectUrl>
-    <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
+    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
```
marklio, line 51: Were you going to set it to true, or remove?

marklio (approved):

Directory.Build.props
Lxiamail: I trie to remove it, but the default value seems false. That's why I set it to true instead.

terrajobst, line 51: FWIW, we mostly set it to false these days. It's just `true` happens to be the default, so we're not entirely consistent. I'd rather we change try-convert 😄

However, for global tools it doesn't really matter `dotnet tool install` seems to ignore it.

terrajobst (approved):

Directory.Build.props
Lxiamail: I see, in that case, I would rather set it back to false to avoid potential prompt from any system.

PR #948, opened 2020-11-06
Update ProxyAuthenticationHandler.cs

brunowarmling:
Fixed a bug with the ProxyAuthenticationHandler that was not handling proxy authentication required properly.

### DIFF
--- a/src/lib/Microsoft.Fx.Portability/Proxy/ProxyAuthenticationHandler.cs
+++ b/src/lib/Microsoft.Fx.Portability/Proxy/ProxyAuthenticationHandler.cs
@@ -46,6 +46,7 @@ namespace Microsoft.Fx.Portability.Proxy
                     {
                         throw new ProxyAuthenticationRequiredException(request.RequestUri);
                     }
+
                     return result;
                 }
                 catch (Exception ex) when (ProxyAuthenticationRequired(ex))
--- a/src/lib/Microsoft.Fx.Portability/Proxy/ProxyAuthenticationHandler.cs
+++ b/src/lib/Microsoft.Fx.Portability/Proxy/ProxyAuthenticationHandler.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using Microsoft.Fx.Portability.Resources;
@@ -33,6 +33,8 @@ namespace Microsoft.Fx.Portability.Proxy
             _proxyProvider = proxyProvider ?? throw new ArgumentNullException(nameof(proxyProvider));
         }
 
+        public HttpStatusCode LastStatusCode { get; private set; }
+
         protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
         {
             int attempts = 0;
@@ -41,7 +43,10 @@ namespace Microsoft.Fx.Portability.Proxy
             {
                 try
                 {
-                    return await base.SendAsync(request, cancellationToken);
+                    HttpResponseMessage result = await base.SendAsync(request, cancellationToken);
+                    this.LastStatusCode = result.StatusCode;
+                    result.EnsureSuccessStatusCode();
+                    return result;
                 }
                 catch (Exception ex) when (ProxyAuthenticationRequired(ex))
                 {
@@ -72,6 +77,14 @@ namespace Microsoft.Fx.Portability.Proxy
             if (ex is ProxyAuthenticationRequiredException)
             {
                 return true;
+            }
+
+            if (ex is HttpRequestException)
+            {
+                if (LastStatusCode == HttpStatusCode.ProxyAuthenticationRequired)
+                {
+                    return true;
+                }
             }
 
 #if FEATURE_WEBEXCEPTION
--- a/src/lib/Microsoft.Fx.Portability/Proxy/ProxyAuthenticationHandler.cs
+++ b/src/lib/Microsoft.Fx.Portability/Proxy/ProxyAuthenticationHandler.cs
@@ -33,8 +33,6 @@ namespace Microsoft.Fx.Portability.Proxy
             _proxyProvider = proxyProvider ?? throw new ArgumentNullException(nameof(proxyProvider));
         }
 
-        public HttpStatusCode LastStatusCode { get; private set; }
-
         protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
         {
             int attempts = 0;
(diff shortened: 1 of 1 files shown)


brunowarmling:
See that the exception "ProxyAuthenticationRequiredException" is being fired after the SendAsync. It means the catch block will never handle the proxy authentication required.

I just download the code in my machine and fixed it. So, i guess that could help people working behind a proxy.

Regards,
Bruno Warmling.

marklio:
Thanks for this contribution! We're looking at this now.

src/lib/Microsoft.Fx.Portability/Proxy/ProxyAuthenticationHandler.cs
```diff
@@ -41,7 +43,10 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
             {
                 try
                 {
-                    return await base.SendAsync(request, cancellationToken);
+                    HttpResponseMessage result = await base.SendAsync(request, cancellationToken);
+                    this.LastStatusCode = result.StatusCode;
```
marklio, line 47: Storing the last status code in a field leaves it open to races between multiple requests. Instead of this pattern, we could throw our own exception that is handled similarly in ProxyAuthenticationRequied().

```suggestion
                    if (this.LastStatusCode == HttpStatusCode.ProxyAuthenticationRequired)
                    {
                        throw SomePrivateExceptionWeHandleBelow();
                    }
```

src/lib/Microsoft.Fx.Portability/Proxy/ProxyAuthenticationHandler.cs
```diff
@@ -41,7 +43,10 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
...
+                    result.EnsureSuccessStatusCode();
```
marklio, line 48: It looks like we don't normally throw here for failing status code, so I'd like to avoid throwing for some condition beyond the proxy situation. This code path has likely regressed over time, and was clearly designed to handle exceptions, but I'd like to not perturb the way it works outside of the proxy failure.

marklio (changes requested):
This looks like a valuable fix. Let's see if we can tighten up the solution.

src/lib/Microsoft.Fx.Portability/Proxy/ProxyAuthenticationHandler.cs
marklio: Let me know if you're not following my suggestion

src/lib/Microsoft.Fx.Portability/Proxy/ProxyAuthenticationHandler.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
```
marklio, line 1: Looks like the build is complaining because of the byte-order mark removal. Let me know if you need help resolving that issue. The build wants this as UTF with a BOM.

src/lib/Microsoft.Fx.Portability/Proxy/ProxyAuthenticationHandler.cs
brunowarmling: Yeah, i got it.

src/lib/Microsoft.Fx.Portability/Proxy/ProxyAuthenticationHandler.cs
brunowarmling: In this part i need help. Maybe i have to do it in a different way...

brunowarmling (commented):
Changed the commit based on your comments

brunowarmling:
Is there something that i have to do?

src/lib/Microsoft.Fx.Portability/Proxy/ProxyAuthenticationHandler.cs
marklio: I can help resolve this tomorrow.

marklio:
> Is there something that i have to do?

Thanks for the update! I like the simplicity of the current PR. I'll get some clarity on why we have a strict byte order mark check in the build. Then, based on that, I'll either fix your BOM, or address the build issue. Thanks again for this contribution!

src/lib/Microsoft.Fx.Portability/Proxy/ProxyAuthenticationHandler.cs
marklio: @brunowarmling, the byte order mark error is not something we can turn off easily. The most straightforward thing would be for you to fix the encoding. Some options for this:
* Open and save in notepad - When you save in notepad, you can specify the encoding in the little box at the bottom. Choose "UTF-8 with BOM".
* Use your editor's functionality for setting the encoding. This depends upon the editor you're using.
  * In VS Code, there is a little option in the status bar to select the encoding before saving.
  * In Visual Studio, there's a drop-down in the save dialog that allows you to save with encoding. When you select this, there is another window that comes up, and you can choose "Unicode (UTF-8 with signature) Codepage 65001".

If you use some other editor, let me know and I'll figure out how to set the encoding. Thanks for helping with this.

brunowarmling: Actually i coded this at visual studio, but changed the file using the browser here in the github... There's a option to do that.
I copied the code from visual studio and pasted at github editor.

When i get home, i will try to use a git client.

brunowarmling: Ok, it's done... I used the GitHub Desktop to send the changes.
But, the build is still failing... i can't figure out why.

marklio: Thanks for your patience here. We're clearly not set up well to give clear feedback on these failures for external PRs. The item failing is another one of the "Style Cop" errors:

> src\lib\Microsoft.Fx.Portability\Proxy\ProxyAuthenticationHandler.cs(48,22): Error SA1513: Closing brace should be followed by blank line

I'm happy to help babysit through any additional errors. Looks like we've got some work to do if we're going to be serious about taking external contributions. Thanks for being the guinea pig here, and sorry for the hassle.

src/lib/Microsoft.Fx.Portability/Proxy/ProxyAuthenticationHandler.cs
```diff
@@ -41,7 +41,12 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
             {
                 try
                 {
-                    return await base.SendAsync(request, cancellationToken);
+                    HttpResponseMessage result = await base.SendAsync(request, cancellationToken);
+                    if (result.StatusCode == HttpStatusCode.ProxyAuthenticationRequired)
+                    {
+                        throw new ProxyAuthenticationRequiredException(request.RequestUri);
+                    }
```
marklio, line 48: Style Cop seems to want a blank line after this (which seems a little silly)

```suggestion
                    }

```

marklio (approved):
Thanks for your patience through this process!

brunowarmling:
> 
> 
> Thanks for your patience through this process!

Thanks for your help and time!

PR #962, opened 2021-03-02
Update nuget feeds

Lxiamail:
Update NuGet Feeds and updated Autfac* components version

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -2,7 +2,7 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="nuget" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
-    <add key="corefx lab" value="https://dotnet.myget.org/F/dotnet-corefxlab/api/v3/index.json" />
+    <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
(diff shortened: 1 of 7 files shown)


NuGet.Config
```diff
@@ -2,7 +2,7 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="nuget" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
-    <add key="corefx lab" value="https://dotnet.myget.org/F/dotnet-corefxlab/api/v3/index.json" />
+    <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
+    <add key="myget-legacy" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/myget-legacy/nuget/v3/index.json" />
```
marklio, line 6: What packages restore from this? I would assume we wouldn't need anything from here.

marklio (approved):

NuGet.Config
Lxiamail: myget-legacy feed is for system.commandline. Looks like argumentsyntax class was removed from newer version of system.commandline package. Using myget-legecy to get the version of system.commandline that it was referencing before. We should move to newer version later, then get rid of myget-legacy

PR #963, opened 2021-03-03
Roll back Autofac verson from 6.0.0 to 4.6.2 since 6.0.0 has problem …

Lxiamail:
…to work with apiport.vsix


marklio (approved):
Looks like some accessibility fixes are in here too. Probably worth making sure the resulting commit includes that in the commit description. I think it's fine though.

Lxiamail:
> Looks like some accessibility fixes are in here too. Probably worth making sure the resulting commit includes that in the commit description. I think it's fine though.

Yes, I forgot the pending PR and commited two accessibility bug fixes for https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1271070/ and https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1270520/

PR #964, opened 2021-03-05
Accessibility bug fixes

Lxiamail:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1263107 
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1270529
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1059540

PR #984, opened 2021-07-01
fix accessibility issue 1059540

Lxiamail:
A11y_[BM].NETPortabilityService_GenerateHTMLFormatPortabilityReport_ScreenReader: Narrator is not announcing the status message after selecting any checkbox under the "Hide Column".
This is an update to PR https://github.com/microsoft/dotnet-apiport/pull/964


marklio (approved):
I assume that these changes have come from someone? If so, I approve. Otherwise, I'm not sure I understand how fieldset works.

PR #989, opened 2021-08-19
update to pool VSEngSS-MicroBuild2019

Lxiamail:


LoopedBard3:
@Lxiamail @marklio Is the new signing error something that should be expected? It looks like the dev branch is currently succeeding on the main repo page so this may mean other updates need to be made elsewhere. If this is expected, I think it looks good. Working on Azure fundamentals stuff so didn't dive too deep into the error.

Lxiamail:
> @Lxiamail @marklio Is the new signing error something that should be expected? It looks like the dev branch is currently succeeding on the main repo page so this may mean other updates need to be made elsewhere. If this is expected, I think it looks good. Working on Azure fundamentals stuff so didn't dive too deep into the error.

Yes, the signing error from DotNet-Apiport-Release-yml pipeline is expected since I created a new pipeline for release from main branch and signing request is still pending. However, I don't expect the   DotNet-Apiport-Release-yml  pipeline to be triggered by PR to dev branch, it whould be only triaged when checkin to main branch. I need to look into it.

LoopedBard3 (approved):
The new pool name looks good and I didn't see any other places where it should need to be changed. Approved 👍 keeping in we should look into why the release yml file was triggered and not the dev one.

